PR #4, opened 2018-03-06
Te 215 dropdown  with icon

matthizou:
Te 215 dropdown  with icon

Add prop to specify an icon for the dropdown list

### DIFF
--- a/src/components/elements/Dropdown/Readme.md
+++ b/src/components/elements/Dropdown/Readme.md
@@ -13,7 +13,20 @@ const { options } = require('./mock-data/options');
 />
 ```
 
-### Content
+### Variations
+
+#### Icon
+An icon can be displayed with the dropdown. [See all available icons](https://react.semantic-ui.com/elements/icon#icon-set)
+
+```jsx
+const { options } = require('./mock-data/options');
+<Dropdown
+  label="Location"
+  options={options}
+  icon="world"
+/>
+```
+
 
 #### Image
 Options can contain images
--- a/src/components/elements/Dropdown/Readme.md
+++ b/src/components/elements/Dropdown/Readme.md
@@ -7,28 +7,22 @@ const { options } = require('./mock-data/options');
 //  { text: 'United Kingdom', value: 'gb' },
 // ];
 
-<Dropdown
-  label="Location"
-  options={options}
-/>
+<Dropdown label="Location" options={options} />;
 ```
 
 ### Variations
 
 #### Icon
-An icon can be displayed with the dropdown. [See all available icons](https://react.semantic-ui.com/elements/icon#icon-set)
+
+The dropdown can contain an icon. [See all available icons](https://react.semantic-ui.com/elements/icon#icon-set)
 
 ```jsx
 const { options } = require('./mock-data/options');
-<Dropdown
-  label="Location"
-  options={options}
-  icon="world"
-/>
+<Dropdown label="Location" options={options} icon="world" />;
 ```
 
-
 #### Image
+
 Options can contain images
 
 ```jsx
@@ -40,8 +34,5 @@ const { optionsWithImages } = require('./mock-data/options');
 //  { text: 'The Black Lodge', value: 'blackLodge', image: '/path/to/image.png' },
 // ];
 
-<Dropdown
-  label="Properties"
-  options={optionsWithImages}
-/>
+<Dropdown label="Properties" options={optionsWithImages} icon="world" />;
 ```
--- a/src/components/elements/Dropdown/component.js
+++ b/src/components/elements/Dropdown/component.js
@@ -47,9 +47,6 @@ export class Component extends PureComponent {
       >
         {icon && <Icon name={icon} />}
         <Dropdown
-          className={getClassNames({
-            dirty: value,
-          })}
           className={value ? 'dirty' : ''}
           defaultValue={defaultValue}
           onChange={this.handleChange}
--- a/src/components/elements/Dropdown/component.js
+++ b/src/components/elements/Dropdown/component.js
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { Dropdown } from 'semantic-ui-react';
+import { Dropdown, Icon } from 'semantic-ui-react';
 import getClassNames from 'classnames';
 
 import { adaptOptions } from './utils/adaptOptions';
@@ -36,21 +36,29 @@ export class Component extends PureComponent {
 
   render() {
     const { value } = this.state;
-    const { label, options } = this.props;
+    const { label, options, icon } = this.props;
     const optionsWithImages = adaptOptions(options);
     const defaultValue = getDefaultValue(optionsWithImages);
     return (
-      <Dropdown
-        className={getClassNames({
-          dirty: value,
-          hasImages: optionsWithImages,
+      <div
+        className={getClassNames('dropdown-container', {
+          'has-images': optionsWithImages,
+          'has-left-icon': icon,
         })}
-        defaultValue={defaultValue}
-        onChange={this.handleChange}
-        options={optionsWithImages || options}
-        placeholder={label}
-        selection
-      />
+      >
+        <Dropdown
+          className={getClassNames({
+            dirty: value,
+          })}
+          className={value ? 'dirty' : ''}
+          defaultValue={defaultValue}
+          onChange={this.handleChange}
+          options={optionsWithImages || options}
+          placeholder={label}
+          selection
+        />
+        {icon && <Icon name={icon} className="left-dropdown-icon" />}
+      </div>
     );
   }
 }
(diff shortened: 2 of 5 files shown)


src/components/elements/Dropdown/component.js
```diff
@@ -35,21 +35,29 @@ export class Component extends PureComponent {
...
     const defaultValue = getDefaultValue(optionsWithImages);
     return (
-      <Dropdown
-        className={getClassNames({
-          dirty: value,
-          hasImages: optionsWithImages,
+      <div
+        className={getClassNames('dropdown-container', {
+          hasImages: !!optionsWithImages,
+          hasLeftIcon: !!icon,
```
matthizou: I am mirroring the case of 'hasImages', but what about always using kebab-case for css classes ('has-left-icon' and 'has-images') ? It would avoid inconsistency

src/components/elements/Dropdown/component.spec.js
```diff
@@ -1,24 +1,39 @@
 import React from 'react';
 import { shallow } from 'enzyme';
+import { Dropdown as SemanticDropdown, Icon } from 'semantic-ui-react';
 
 import { Component as Dropdown } from './component';
 
-const options = [{ text: 'someText', value: 'someValue' }];
-const optionsWithImages = [
+const OPTIONS = [{ text: 'someText', value: 'someValue' }];
+const OPTIONS_WITH_IMAGES = [
```
matthizou: I apply another convention that I strongly support here: using capital letters for mock data - to make it obvious that those objects shouldn't be changed.
Too often those reused objects are modified in a test, screwing up the following tests.

src/components/elements/Dropdown/Readme.md
```diff
@@ -13,7 +13,20 @@ const { options } = require('./mock-data/options');
 />
 ```
 
-### Content
+### Variations
+
+#### Icon
+An icon can be displayed with the dropdown. [See all available icons](https://react.semantic-ui.com/elements/icon#icon-set)
```
marcdavi-es: Can we be more specific? Maybe

`The dropdown trigger can contain an icon`

src/components/elements/Dropdown/component.js
marcdavi-es: 1. I don't think you need to coerce to false here... `classnames` does that for you.
2. Change to kebab case if you want.

src/components/elements/Dropdown/component.js
```diff
@@ -59,9 +67,12 @@ Component.displayName = 'Dropdown';
 Component.defaultProps = {
   label: '',
   onChange: Function.prototype,
+  icon: '',
```
marcdavi-es: I think `null` is a better default value here.

src/components/elements/Dropdown/component.js
```diff
@@ -35,21 +35,29 @@ export class Component extends PureComponent {
...
+            dirty: value,
+          })}
+          className={value ? 'dirty' : ''}
+          defaultValue={defaultValue}
+          onChange={this.handleChange}
+          options={optionsWithImages || options}
+          placeholder={label}
+          selection
+        />
+        {icon && <Icon name={icon} className="left-dropdown-icon" />}
```
marcdavi-es: I think we can make this easier by:
1. losing the wrapping `div`
2. Passing something like `<span>{icon && <Icon name={icon} />}{label}</span>` as the placeholder.

src/semantic/src/themes/livingstone/modules/dropdown.overrides
```diff
@@ -63,8 +63,30 @@
   content: '\f0d7';
 }
 
-.ui.dropdown.hasImages > .dropdown.icon {
-  padding-top: @dropdownWithImagesCaretPaddingTop;
+.dropdown-container{
```
marcdavi-es: I think we can lose all of this if we follow the `props.placeholder` route.

We should avoid adding extra classes wherever possible. In the case of `.hasImages`, there was no way around it.

src/components/elements/Dropdown/component.js
marcdavi-es: screen shot 2018-03-06 at 19 51 37

...gets us most of the way there, no?

matthizou: Nice trick !! I wish I had known it before.
We'll lose the icon once the value is set, though, or even when the menu is expanded. I personally don't mind, but consumers of our components may -- the expected behaviour for an dropdown icon is to stick around all the time (i.e the one from Semantic UI).

marcdavi-es: Is that the expected behaviour? Check with Julio.

matthizou: ok, checked - he prefers the icon to be there all the time.

src/components/elements/Dropdown/Readme.md
matthizou: This comment can be ignored if we don't use the placeholder technique, right ?

src/semantic/src/themes/livingstone/modules/dropdown.overrides
matthizou: Agreed. I tried to keep it to a minimum here, and to reuse variables when possible.
This comment can be ignored if we don't use the placeholder route, right ?

src/semantic/src/themes/livingstone/modules/dropdown.variables
```diff
@@ -3,6 +3,7 @@
 // .selection.dropdown
 
 @selectionHoverBorderColor          : @inputHoverBorderColor;
+@selectionBorderRadius              : 0;
```
matthizou: Following a comment from Julio

src/components/elements/Dropdown/Readme.md
marcdavi-es: Maybe `The dropdown can contain an icon` ?

src/components/elements/Dropdown/component.js
marcdavi-es: OK fair enough. 

I think we can simplify this in any case. Less moving parts...

```jsx
// component.js
      <div className="dropdown-container">
        {icon && <Icon name={icon} />}
        <Dropdown
          ...
        />
      </div>
```

```css
// dropdown.overrides
.dropdown-container {
  position: relative;

  & > i.icon {
    position: absolute;
    left: 5px;
    top: @selectionVerticalPadding;
    z-index: @selectionZIndex + 1;
    color: @itemColor;

     & + .ui.dropdown {
        padding-left: @selectionIconDistance;
     }
  }

}
```

src/components/elements/Dropdown/Readme.md
matthizou: Sure

src/components/elements/Dropdown/component.js
```diff
@@ -35,21 +35,28 @@ export class Component extends PureComponent {
...
+          'has-images': optionsWithImages,
         })}
-        defaultValue={defaultValue}
-        onChange={this.handleChange}
-        options={optionsWithImages || options}
-        placeholder={label}
-        selection
-      />
+      >
+        {icon && <Icon name={icon} />}
```
matthizou: As per previous CR suggestion, the extra classes "left-icon" and "has-left-icon" were removed
'has-images' unfortunately cannot, as the added images are deeply nested inside the 2nd child of the container

marcdavi-es: ...but `has-images` can stay on the `Dropdown` component, right?

It doesn't need to be change.

src/components/elements/Dropdown/component.js
```diff
@@ -59,9 +66,12 @@ Component.displayName = 'Dropdown';
 Component.defaultProps = {
   label: '',
   onChange: Function.prototype,
+  icon: null,
 };
 
 Component.propTypes = {
+  /** Icon for the dropdown */
```
marcdavi-es: `An icon to display in the dropdown`

src/components/elements/Dropdown/component.js
matthizou: Maybe there some css magic I don't know, but given two siblings elements, A and B, can we style A depending on a class B has ? 
The other way around is simple - the " + " selector, that selects the next sibling. But there is nothing to do it the other way around

marcdavi-es: OK - because we need to add the extra `top` distance for dropdowns which have both images and icons.

src/components/elements/Dropdown/component.js
```diff
@@ -36,21 +36,28 @@ export class Component extends PureComponent {
...
-        placeholder={label}
-        selection
-      />
+      >
+        {icon && <Icon name={icon} />}
+        <Dropdown
+          className={getClassNames({
+            dirty: value,
+          })}
+          className={value ? 'dirty' : ''}
```
marcdavi-es: `className` is set twice here.

matthizou: Good catch

PR #5, opened 2018-03-06
TE-212 Form collection

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-212)

Creates component, tests and examples for the Form collection,.

### Any other notes
I had to make `name` and `onChange` props consistent across all input elements.


src/components/collections/Form/component.js
```diff
@@ -0,0 +1,94 @@
...
+  /** The child components and elements. */
+  children: PropTypes.node.isRequired,
+  /** The text to display as a heading at the top of the form. */
+  headingText: PropTypes.string,
+  /** The function to call when the form is submitted
+   *  @param {Object} values - The values of the inputs in the form.
+   */
+  onSubmit: PropTypes.func,
+  /** The text to display on the primary call to action. */
+  primaryCTAText: PropTypes.string.isRequired,
```
matthizou: "submitButtonText". Clearer, and that is what it is.
This acronym "CTA" may (will) be obscure for many -- and the term "submit" is already used ("onSubmit").

src/components/collections/Form/component.js
```diff
@@ -0,0 +1,94 @@
+import React, { PureComponent, Children } from 'react';
```
matthizou: Comments for the file:
What about removing secondaryCTA, it is just a link that the user can easily add himself, where he wants, if he does.
Similarly, I would also let the consumer add its own <Button type="submit">, to get more flexibility about its location and props. Pretty much like Semantic UI and Bootstrap react are doing. 
Shorthands look convenient, but quickly, when designing new forms, their rigidity turns against us

src/components/collections/Form/component.spec.js
```diff
@@ -0,0 +1,183 @@
...
+const textInputChild = <TextInput name="someName" />;
+const primaryCTAText = 'somePrimaryCTAText';
+const headingText = '👥';
+
+describe('<Form />', () => {
+  it('should render a single Semantic UI `Card` component', () => {
+    const form = shallow(
+      <Form primaryCTAText={primaryCTAText}>{stringChildren}</Form>
+    );
+    const actual = form.find('Card');
```
matthizou: `import` { Card as SemanticCard } from 'semantic-ui';
//.....
const actual = form.find(SemanticCard);

src/components/collections/Form/component.js
```diff
@@ -0,0 +1,94 @@
...
+              children,
+              child =>
+                child.type === Form.Group
+                  ? this.cloneFormGroup(child)
+                  : this.cloneInput(child)
+            )}
+            {secondaryCTA && (
+              <a onClick={secondaryCTA.onClick}>{secondaryCTA.text}</a>
+            )}
+            <Button isPositionedRight>{primaryCTAText}</Button>
```
matthizou: type="submit"

src/components/collections/Form/component.js
marcdavi-es: I'll give optional shorthands for `submitButton` and `secondaryAction` so that it's super easy to consume the API.

If the user wants to compose their own, they are able to do that with links, buttons, whatever they want. There's nothing stopping that.

src/components/collections/Form/component.js
marcdavi-es: I'm not sure that's necessary?

src/components/collections/Form/component.spec.js
marcdavi-es: Why is that better?

src/components/collections/Form/component.js
matthizou: It is not necessary, because the default type of a button is "submit".
It is purely semantic. Up to you

src/components/collections/Form/component.js
matthizou: Cool.
What about calling "secondaryAction" , "actionLink" ?

src/components/collections/Form/component.spec.js
matthizou: Slighlty better imo,  -- as our component is already called 'Card', reading the line `const actual = form.find(SemanticCard);` require 0% context to understand what is being done
Tradeoff is the extra import. Up to you

src/components/collections/Form/component.js
marcdavi-es: I don't think we need it.

PR #7, opened 2018-03-07
TE-219 OwnerSignUp widget

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-219)

Create component, tests and examples for OwnerSignUp

### DIFF
--- a/src/components/widgets/OwnerSignUp/component.js
+++ b/src/components/widgets/OwnerSignUp/component.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Form } from '../../collections/Form';
+import { TextInput } from '../../elements/TextInput';
+
+export const Component = ({ onSubmit }) => (
+  <Form
+    headingText="Owner Signup"
+    onSubmit={onSubmit}
+    submitButtonText="Sign up"
+  >
+    <TextInput label="First Name" name="firstName" />
+    <TextInput label="Last Name" name="lastName" />
+    <TextInput label="Email" name="email" />
+  </Form>
+);
+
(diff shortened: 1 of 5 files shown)


src/components/widgets/OwnerSignUp/component.js
```diff
@@ -0,0 +1,30 @@
...
+    <TextInput label="First Name" name="firstName" />
+    <TextInput label="Last Name" name="lastName" />
+    <TextInput label="Email" name="email" />
+  </Form>
+);
+
+Component.displayName = 'OwnerSignUp';
+
+Component.defaultProps = {
+  onSubmit: Function.prototype,
```
matthizou: Never seen that before, using the function prototype as an empty function ( () => {} ).  But yes, why not. Interesting

src/components/widgets/OwnerSignUp/component.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Form } from '../../collections/Form';
+import { TextInput } from '../../elements/TextInput';
+
+export const Component = ({ onSubmit }) => (
```
matthizou: Missing description

src/components/widgets/OwnerSignUp/component.js
marcdavi-es: I think this is better for a noop. The empty function: 
1. creates a new reference which we don't need
2. gets picked up by test coverage

src/components/widgets/OwnerSignUp/component.js
marcdavi-es: Fixed

PR #12, opened 2018-03-07
TE-225-Styleguidist minor config change

matthizou:
TE-225-Styleguidist minor config change

Changing the way the import is display
image


marcdavi-es:
I don't think we want to do this until we know what the external api looks like.

PR #21, opened 2018-03-14
TE-216 CaptchaInput

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-216)

Creates CaptchaInput component, tests and examples.


styleguide/webpack.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
         use: ['style-loader', 'css-loader'],
       },
       {
-        test: /\.(eot|svg|ttf|woff|woff2|png)$/,
+        test: /\.(eot|svg|ttf|woff|woff2|png|jpeg)$/,
```
davidpelayo: maybe `jpeg` extension to match optionally the `e` so both `jpeg` and `jpg` are accepted?

marcdavi-es: Totally see what you're saying - Right now I'm just adding only what we need.

PR #22, opened 2018-03-15
TE-234 SearchBar

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-234)

Creates component, tests and examples for SearchBar

### Any other notes
Required a couple of consequent changes to existing styles.

### DIFF
--- a/src/semantic/src/themes/livingstone/elements/button.variables
+++ b/src/semantic/src/themes/livingstone/elements/button.variables
@@ -14,3 +14,5 @@
 @disabledBackgroundColor: @greyFive;
 @disabledFontColor: @pureWhite;
 @disabledOpacity: 1;
+
+@iconMargin: 0em @iconDistance 0em -(@iconDistance);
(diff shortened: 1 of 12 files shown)


src/semantic/src/themes/livingstone/elements/button.variables
```diff
@@ -14,3 +14,5 @@
 @disabledBackgroundColor: @greyFive;
 @disabledFontColor: @pureWhite;
 @disabledOpacity: 1;
+
+@iconMargin: 0em @iconDistance 0em -(@iconDistance);
```
davidpelayo: 0 values don't need units.

src/semantic/src/themes/livingstone/modules/dropdown.variables
```diff
@@ -2,6 +2,10 @@
 
 // .selection.dropdown
 
+@selectionMaxWidth                  : 350px;
+@selectionMinWidth                  : unset;
```
davidpelayo: IE11 doesn't like `unset`. A fallback needs to be provided.

PR #23, opened 2018-03-16
TE-229 Sprint six style fixes

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-229)

Makes a number of fixes carried over from sprint six.

### DIFF
--- a/src/components/elements/Submenu/adaptOptions.js
+++ b/src/components/elements/Submenu/adaptOptions.js
@@ -1,0 +1,7 @@
+/**
+ * Adds unique key to each option in an array.
+ * @param  {Object[]} options
+ * @return {Object[]}
+ */
+export const adaptOptions = options =>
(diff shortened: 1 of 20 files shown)


src/components/elements/Submenu/adaptOptions.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Adds unique key to each option in an array.
+ * @param  {Object[]} options
+ * @return {Object[]}
+ */
+export const adaptOptions = options =>
+  options.map(option => ({ key: option.text, ...option }));
```
davidpelayo: How is so? Is it `option.text` unique?

marcdavi-es: Good question. Technically it doesn't have to be unique. BUT practically if the consumer is using two identical strings in the same submenu, they are nuts.

PR #26, opened 2018-03-20
TE-238 Container

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-238)

- A simple `<Container />` widget in which multiple widgets could be rendered on a same column.

### Additional Notes
- The decision and definition of subset of Semantic `<Grid />`/`<Grid.Column />`/`<Grid.Row />` props wanted to be reused will be implemented in [TE-255](https://youtrack.lodgify.net/issue/TE-255).


src/components/widgets/Container/component.js
```diff
@@ -0,0 +1,29 @@
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+import { Grid, Segment } from 'semantic-ui-react';
+
+/**
+ * A container which renders widgets inside a single column
+ */
+export const Component = ({ children }) => (
+  <Grid columns={1}>
```
marcdavi-es: Presumably `columns` will be controlled by props?

src/components/widgets/Container/component.js
```diff
@@ -0,0 +1,29 @@
...
+    <Grid.Row>
+      <Grid.Column>
+        {Children.map(children, (child, index) => (
+          <Segment key={index}>{child}</Segment>
```
marcdavi-es: Is the segment necessary? Shouldn't each child live directly in a column?

davidpelayo: Yes, you're right.

src/components/widgets/Container/component.js
davidpelayo: I don't know what to say here. I believe we must stack widgets vertically in a single column.
How widgets get displayed across columns and rows should be a responsibility of the Template Grid renderer.

marcdavi-es: Yep the template grid thing does the logical bit.

But then it needs a dumb component which it can tell to occupy a certain number of columns, right?

src/components/collections/Grid/GridColumn/component.js
```diff
@@ -0,0 +1,30 @@
...
+ * Semantic UI Grid.Column.
+ *
+ * The [Semantic UI props](https://react.semantic-ui.com/collections/grid) can be for this component.
+ * @type {Object}
+ */
+export const Component = props => (
+  <Grid.Column
+    {...Object.assign({}, props, {
+      width: getTwelveColumnsWidthValue(props.width),
+    })}
```
marcdavi-es: I think we tend to use object spread in other places?

`{...{ ...props, width: getTwelveColumnsWidthValue(props.width) }}`

src/components/collections/Grid/GridColumn/component.js
```diff
@@ -0,0 +1,30 @@
...
+  />
+);
+
+Component.defaultProps = {
+  /** Value between 1 and 12 */
+  width: 1,
+};
+
+Component.propTypes = {
+  width: PropTypes.number,
```
marcdavi-es: I think I used `string` elsewhere, as Semantic allows either `1`, `'1'` or `'one'`. 

Probably we're both too loose and we should follow the Semantic approach of using `PropTypes.oneOf([ ...all number, string numeral and full strings values between one and twelve ])` ?

screen shot 2018-03-26 at 11 13 17

src/components/collections/Grid/GridRow/component.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Grid } from 'semantic-ui-react';
+
+import { DEFAULT_MAX_WIDTH } from '../constants';
+import { getTwelveColumnsWidthValue } from '../utils';
+/**
+ * GridRow is the Lodgify UI 12-columns based component for the
+ * Semantic UI Grid.Row.
+ *
+ * The [Semantic UI props](https://react.semantic-ui.com/collections/grid) can be for this component.
```
marcdavi-es: I think we should default to 'masking' the Semantic UI implementations, rather than referring to them. Is there a subset of the `Grid.Row` props which we want to expose in `GridRow`? We should be specific here.

src/components/collections/Grid/GridRow/component.js
```diff
@@ -0,0 +1,21 @@
...
+ * @type {Object}
+ */
+export const Component = props => (
+  <Grid.Row
+    {...Object.assign({}, props, {
+      columns: getTwelveColumnsWidthValue(DEFAULT_MAX_WIDTH),
+    })}
```
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/26/files#r177023294

src/components/collections/Grid/component.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Grid } from 'semantic-ui-react';
+
+import { DEFAULT_MAX_WIDTH } from './constants';
+import { getTwelveColumnsWidthValue } from './utils';
+/**
+ * Grid is the Lodgify UI 12-columns based component for the
+ * Semantic UI Grid.
+ *
+ * The [Semantic UI props](https://react.semantic-ui.com/collections/grid) can be for this component.
```
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/26/files#r177025041

src/components/collections/Grid/component.js
```diff
@@ -0,0 +1,21 @@
...
+ * @type {Object}
+ */
+export const Component = props => (
+  <Grid
+    {...Object.assign({}, props, {
+      columns: getTwelveColumnsWidthValue(DEFAULT_MAX_WIDTH),
+    })}
```
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/26/files#r177023294

src/components/collections/Grid/GridColumn/component.js
```diff
@@ -0,0 +1,30 @@
...
+import PropTypes from 'prop-types';
+import { Grid } from 'semantic-ui-react';
+
+import { getTwelveColumnsWidthValue } from '../utils';
+/**
+ * GridColumn is the Lodgify UI 12-columns based component for the
+ * Semantic UI Grid.Column.
+ *
+ * The [Semantic UI props](https://react.semantic-ui.com/collections/grid) can be for this component.
+ * @type {Object}
```
marcdavi-es: `@returns {Object}`

src/components/collections/Grid/GridColumn/component.js
davidpelayo: I'd keep it as simple as possible, to be honest. Which means, I would go for a number based definition only.

src/components/collections/Grid/GridRow/component.js
```diff
@@ -0,0 +1,21 @@
...
+import { Grid } from 'semantic-ui-react';
+
+import { DEFAULT_MAX_WIDTH } from '../constants';
+import { getTwelveColumnsWidthValue } from '../utils';
+/**
+ * GridRow is the Lodgify UI 12-columns based component for the
+ * Semantic UI Grid.Row.
+ *
+ * The [Semantic UI props](https://react.semantic-ui.com/collections/grid) can be for this component.
+ * @type {Object}
```
marcdavi-es: https://github.com/lodgify/lodgify-ui/pull/26/files#r177025419

src/components/collections/Grid/component.js
```diff
@@ -0,0 +1,21 @@
...
+import { Grid } from 'semantic-ui-react';
+
+import { DEFAULT_MAX_WIDTH } from './constants';
+import { getTwelveColumnsWidthValue } from './utils';
+/**
+ * Grid is the Lodgify UI 12-columns based component for the
+ * Semantic UI Grid.
+ *
+ * The [Semantic UI props](https://react.semantic-ui.com/collections/grid) can be for this component.
+ * @type {Object}
```
marcdavi-es: https://github.com/lodgify/lodgify-ui/pull/26/files#r177025419

src/components/collections/Grid/GridRow/component.js
davidpelayo: Yes. Agreed.

src/components/collections/Grid/index.js
```diff
@@ -0,0 +1,5 @@
+export { Component as Grid } from './component';
+
+export { COLUMNS, SEMANTIC_FULL_WIDTH_COLUMN } from './constants';
```
marcdavi-es: Do we need to expose `SEMANTIC_FULL_WIDTH_COLUMN`

src/components/collections/Grid/utils/getTwelveColumnsWidthValue.js
```diff
@@ -0,0 +1,24 @@
+import { isNumber } from 'lodash';
+
+import {
+  SEMANTIC_FULL_WIDTH_COLUMN,
+  COLUMNS,
+  DEFAULT_COLUMN_WIDTH,
+  DEFAULT_MAX_WIDTH,
+} from '../constants';
+
+export const getTwelveColumnsWidthValue = given12BasedWidth => {
```
marcdavi-es: I think we can avoid this complication by hardwiring `columns="twelve"` into all our `Grid`s and `GridRow`s ? Then Semantic does the work for us?

screen shot 2018-03-26 at 11 19 53

src/components/collections/Grid/index.js
davidpelayo: Oooops, totally right. Not outside the Grid concerns.

src/components/widgets/Container/component.js
```diff
@@ -0,0 +1,36 @@
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+
+import { Grid } from 'collections/Grid';
+import { GridRow } from 'collections/Grid/GridRow';
+import { GridColumn } from 'collections/Grid/GridColumn';
+
+/**
+ * A container which renders widgets inside of it.
+ * By default spans to 100% width stacking widgets vertically.
```
marcdavi-es: `@returns {Object}`

src/components/collections/Grid/utils/getTwelveColumnsWidthValue.js
davidpelayo: Yeah, but we would still need this for the Columns, though.

marcdavi-es: How come?

davidpelayo:
The decision and definition of subset of Semantic Grid/Grid.Column/Grid.Row props wanted to be reused will be implemented in a separate ticket.

PR #27, opened 2018-03-20
TE-239 Image

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-239)

- The Image widget.
- Lazy loading or [react-graceful-image](https://github.com/linasmnew/react-graceful-image) similar techniques will be implemented later.

### DIFF
--- a/src/components/widgets/Image/Readme.md
+++ b/src/components/widgets/Image/Readme.md
@@ -1,0 +1,16 @@
+### Usage
+
+#### No `imageUrl` specified
+
+```jsx
+<Image />
+```
+
+#### Standard usage
+
+```jsx
+<Image
+  imageUrl="//li3.cdbcdn.com/oh/522a12d9-ab51-4635-94c1-42536f286e4d.jpg?w=1140&mode=max&f=32"
+  alternativeText='Hey'
+/>
+```
--- a/src/components/widgets/Image/Readme.md
+++ b/src/components/widgets/Image/Readme.md
@@ -1,17 +1,21 @@
 ### Usage
 
-#### No `imageUrl` specified
-
-```jsx
-<Image />
-```
-
 #### Standard Example
 
+Fluid by default
 ```jsx
 <Image
   imageUrl="//li3.cdbcdn.com/oh/522a12d9-ab51-4635-94c1-42536f286e4d.jpg?w=1140&mode=max"
-  alternativeText='Seamlessly image loading amuses me'
+  alternativeText='Seamlessly fluid image loading amuses me'
+/>
+```
+
+Not fluid image
+```jsx
+<Image
+  imageUrl="//li3.cdbcdn.com/oh/522a12d9-ab51-4635-94c1-42536f286e4d.jpg?w=320&mode=max"
+  isFluid={false}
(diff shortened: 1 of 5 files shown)


src/components/widgets/Image/component.js
```diff
@@ -0,0 +1,91 @@
...
+        <Image
+          src={imageUrl}
+          alt={alternativeText}
+          className={cx(LEGACY_CLASSNAME, className)}
+          title={imageTitle}
+          fluid
+        >
+          <Label content="Image not found!" icon="warning" />
+        </Image>
+      </picture>
```
davidpelayo: I prefer to have this encapsulated instead of repeated. But, as it's described, it's just an initial proposal.

marcdavi-es: You mean deduplicate all the other stuff and do `!imageUrl && <Label content="Image not found!" icon="warning" />` or something?

davidpelayo: If you see the code:
```
{!!imageUrl.length && ( .... )}
{!imageUrl.length && ( ... )}
```
In order to cover both cases. I believe we should simplify this logic. However, we should stablish the boundaries on what showing the “Image not found” label means. Does it really mean not having `imageUrl` props informed or even having it but not being reachable / HTTP-resolvable?
--
copying this comment to the PR for the sake of the discussion traceability.

src/components/widgets/Image/component.js
```diff
@@ -0,0 +1,77 @@
...
+  imageTitle,
+  className,
+  onLoad,
+  isFluid,
+}) => (
+  <picture role="figure">
+    {sources.map(({ srcset, media }, index) => (
+      <source srcSet={srcset} media={media} key={index} />
+    ))}
+    {React.createElement(Image, {
```
marcdavi-es: Why don't we just use JSX here?

davidpelayo: Because the output of my implementation differs from this one:
```jsx
<Image>
  {!imageUrl && <Label />}
</Image>
```

src/components/widgets/Image/Readme.md
```diff
@@ -0,0 +1,44 @@
+### Usage
```
marcdavi-es: This examples files doesn't really fit the pattern we have for others. What about

````
```
<Image src="//li3.cdbcdn.com/oh/522a12d9-ab51-4635-94c1-42536f286e4d.jpg?w=1140&mode=max" />
```

### Variations

#### Fluid

An image fills the width of its containing element by default

```jsx
<Image
  imageUrl="//li3.cdbcdn.com/oh/522a12d9-ab51-4635-94c1-42536f286e4d.jpg?w=1140&mode=max"
  alternativeText='Seamlessly fluid image loading amuses me'
/>
<Image
  imageUrl="//li3.cdbcdn.com/oh/522a12d9-ab51-4635-94c1-42536f286e4d.jpg?w=320&mode=max"
  isFluid={false}
  alternativeText='Seamlessly non-fluid image loading amuses me'
/>
```

#### Responsive 

```
...
```

#### No url

```
...
```
````

What do you think? Sorry to be picky.

Just noticed fluid example will need a wrapping div and Divider between.

PR #28, opened 2018-03-20
TE-213 PhoneInput

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-213)

Adds PhoneInput component, tests and examples.

### Any other notes
`props.icon` is a minor addition to InputController.

phone-input

### DIFF
--- a/src/components/elements/InputController/utils/getValue.js
+++ b/src/components/elements/InputController/utils/getValue.js
@@ -4,5 +4,7 @@ import { get } from 'lodash';
  * @param  {Object|any} eventOrValue
  * @return {any}
  */
-export const getValue = eventOrValue =>
-  get(eventOrValue, ['target', 'value']) || eventOrValue;
+export const getValue = eventOrValue => {
+  const nestedValue = get(eventOrValue, ['target', 'value']);
+  return nestedValue !== undefined ? nestedValue : eventOrValue;
+};
--- a/src/components/elements/InputController/utils/getValue.js
+++ b/src/components/elements/InputController/utils/getValue.js
@@ -1,4 +1,4 @@
-import { get } from 'lodash';
+import { get, isUndefined } from 'lodash';
 
 /**
  * @param  {Object|any} eventOrValue
@@ -6,5 +6,5 @@ import { get } from 'lodash';
  */
 export const getValue = eventOrValue => {
   const nestedValue = get(eventOrValue, ['target', 'value']);
-  return nestedValue !== undefined ? nestedValue : eventOrValue;
+  return !isUndefined(nestedValue) ? nestedValue : eventOrValue;
 };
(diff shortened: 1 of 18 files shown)


src/components/elements/InputController/utils/getValue.js
```diff
@@ -4,5 +4,7 @@ import { get } from 'lodash';
  * @param  {Object|any} eventOrValue
  * @return {any}
  */
-export const getValue = eventOrValue =>
-  get(eventOrValue, ['target', 'value']) || eventOrValue;
+export const getValue = eventOrValue => {
+  const nestedValue = get(eventOrValue, ['target', 'value']);
+  return nestedValue !== undefined ? nestedValue : eventOrValue;
```
davidpelayo: For a matter of consistency maybe using `isUndefined(nestedValue)` from [lodash](https://github.com/lodash/lodash/blob/4.17.5/lodash.js#L12189)?

src/components/elements/PhoneInput/utils/getIconOrFlag.js
```diff
@@ -0,0 +1,11 @@
...
+import { Flag, Icon } from 'semantic-ui-react';
+
+/**
+ * If country code known, return an appropriate flag component.
+ * Default to returning a phone icon.
+ * @param  {String} country
+ * @return {Object}
+ */
+export const getIconOrFlag = country =>
+  country ? <Flag name={country.toLowerCase()} /> : <Icon name="phone" />;
```
davidpelayo: `isString(country)` and flag code validity check is needed here for the first part of the ternary.

src/components/elements/PhoneInput/utils/parseValue.js
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Parse country code and number format from value.
+ * @param  {String} value
+ * @return {Object}
+ */
+export const parseValue = value => {
+  const { country, phone } = parse(value);
+  const formattedValue = phone && format(phone, country, 'International');
+  return {
+    country,
```
davidpelayo: I wonder whether we could customize a default country in case `parse` doesn't work as expected given a weird, empty, unknown or malformed value?

marcdavi-es: We can expose a `defaultCountry` prop. Right now I'm not sure what the user need is though? Do you have an idea from the existing implementation?

src/components/elements/PhoneInput/utils/getIconOrFlag.js
marcdavi-es: 1. I don't think we need string validation. The function is documented as accepting a string AND it only takes the output of `libphonenumber-js` `parse`, which can only be undefined or a string.
2. Yes - we do need to validate the flag name.

PR #30, opened 2018-03-22
TE-233 Tidy up Semantic files

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-233)

Follows the structure of the default variables files in livingstone variables and overrides files.


marcdavi-es:
Stuck on https://youtrack.lodgify.net/issue/TE-252

PR #33, opened 2018-03-23
TE-217 Contact widget

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-217)

Adds component, tests and examples for Contact widget.

### Any other notes
- Exposed a `width` prop for children of `InputGroup` when rendered in a `Form`. Made a note to add this to `Form` examples here https://youtrack.lodgify.net/issue/TE-217
- Made a minor fix to `CaptchaInput` to make sure the `TextInput` has the right padding.


src/components/widgets/Contact/component.js
```diff
@@ -0,0 +1,98 @@
...
+ * @returns {Object}
+ */
+export const Component = ({
+  captchaInputImage,
+  guestsOptions,
+  onSubmit,
+  propertyOptions,
+  roomOptions,
+}) => (
+  <Form headingText="Contact" onSubmit={onSubmit} submitButtonText="Send">
```
davidpelayo: These texts as props maybe better?

src/components/collections/Form/getInputWidth.js
```diff
@@ -0,0 +1,8 @@
+import { get } from 'lodash';
+
+/**
+ * Get the width prop from the input.
+ * @param  {String|Object} input
+ * @return {String|undefined}
+ */
+export const getInputWidth = input => get(input, ['props', 'width']);
```
davidpelayo: Is there any reason why you prefer to introduce props width as array instead of a string dot based selection like `props.width`?

marcdavi-es: Just following the signature for `lodash.get`

This defends against cases when a string is passed as a child to `Form`. A string child doesn't have props so this function will just return `undefined`.

I think this is tidier than

`input => input.props && input.props.width`

?

src/components/widgets/Contact/component.js
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/32#discussion_r177018724

PR #36, opened 2018-03-27
TE-242 FeaturedProperty widget

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-242)

Creates component, examples and tests for FeaturedProperty


src/components/widgets/FeaturedProperty/Readme.md
```diff
@@ -0,0 +1,14 @@
+```jsx
+const { twoCats } = require('./mock-data/images');
```
davidpelayo: The only downside of this method is the end user doesn't actually see the code behind the scenes

src/components/widgets/FeaturedProperty/component.js
```diff
@@ -0,0 +1,74 @@
...
+  imageUrl,
+  guestsNumber,
+  locationName,
+  nightPrice,
+  onClick,
+  propertyName,
+  propertyType,
+  ratingNumber,
+}) => (
+  <Card as="div" onClick={onClick}>
```
davidpelayo: A doubt incoming in my mind.
For the matter of logging/tracking services, I think we should dispatch user action events upwards just after processing the event first.
This means capturing this `onClick`, doing something with it (meaning a middle-man handler) and dispatching it to `props.onClick`.
What do you think?

src/semantic/src/themes/livingstone/views/card.overrides
```diff
@@ -0,0 +1,93 @@
...
+
+/* Icons */
+
+/* Links */
+
+/* Description */
+  .description {
+    margin-top: @descriptionDistance;
+
+    h5.ui.header {
```
davidpelayo: Why is there the need of `h5.ui.header`?

src/components/widgets/FeaturedProperty/component.js
```diff
@@ -0,0 +1,74 @@
...
+    <Image src={imageUrl} />
```
davidpelayo: Alternative texts are always good for 🙈 people

src/components/widgets/FeaturedProperty/Readme.md
marcdavi-es: Yes - I think we need to make our approach to mocking more consistent. Probably a separate lump of work?

src/semantic/src/themes/livingstone/views/card.overrides
marcdavi-es: For the price :)

screen shot 2018-03-27 at 15 27 32

src/components/widgets/FeaturedProperty/component.js
davidpelayo: I think we can remove it for now. However, have please a look at [this](https://github.com/lodgify/lodgify-ui/pull/36#pullrequestreview-107291187).

src/components/widgets/FeaturedProperty/Readme.md
davidpelayo: Yes, but let's forget it for now.

src/components/widgets/FeaturedProperty/component.js
marcdavi-es: I'll take it out.

PR #37, opened 2018-03-27
TE-241 Video widget

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-241)

- Defines the Video widget.


src/components/widgets/Video/component.js
```diff
@@ -0,0 +1,27 @@
...
+
+Component.defaultProps = {
+  isUrl: false,
+  videoInput: null,
+};
+
+Component.propTypes = {
+  /** Whether the input is an url or not
+   */
+  isUrl: PropTypes.bool,
```
marcdavi-es: I don't think we need this prop if we're validating the string?

If it's a url, do the url dance, no?

src/components/widgets/Video/component.js
```diff
@@ -0,0 +1,27 @@
...
+  /** The string that will be used to build the video player. It could be
+   *  whether a URL or a embeddable HTML snippet.
+   */
+  videoSource: PropTypes.string,
```
marcdavi-es: `videoInput`

src/components/widgets/Video/utils/getVideoContent.js
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <div
+        className="video is-html"
+        dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(videoInput) }}
+      />
+    );
+  }
+
+  // Otherwise
+  return null;
```
marcdavi-es: I don't know if this should fail silently. Perhaps we should throw an error if the `videoInput` is not valid.

davidpelayo: I wouldn't get now into error handling. But surely this is not a good pattern.

davidpelayo: Anyway, makes sense. I will throw an error instead.

src/components/widgets/Video/component.js
marcdavi-es: Ah I see. I think it's OK though. If the consumer is passing a bad string, they can eat the performance hit 🍔

src/components/widgets/Video/component.js
davidpelayo: I renamed all to `videoSource`. Do you think is more understandable `videoInput`?

marcdavi-es: I prefer `videoSource`

marcdavi-es: but when I looked before they were inconsistent?

davidpelayo: Yes, now they are fixed.

PR #42, opened 2018-04-03
TE-265 FullBleed component

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-265)

Creates a FullBleed component to fill width of container and height of browser viewport.

### Any other notes
I made a 'Layout' category in the Lodgify UI styleguide and put FullBleed in there with the Grid stuff.

screen shot 2018-04-03 at 15 17 32


src/components/layout/FullBleed/component.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * A full bleed takes up the width of its containing element
+ * and the height of the viewport.
+ *
+ * @returns {Object}
+ */
+export const Component = ({ children, imageUrl }) => (
+  <Segment
+    className="full-bleed"
+    style={{ backgroundImage: getBackgroundImageUrl(imageUrl) }}
```
davidpelayo: Is there any specific reason why you went for the backgroundImage strategy?

marcdavi-es: Si - I think that using this with the `background-size: cover` and `background-position: center` css statements is the simplest way of making sure the image displays well across devices.

PR #43, opened 2018-04-03
TE-265 Image widget label

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-265)

Adds a `label` prop to the Image component.


src/components/typography/Paragraph/index.js
```diff
@@ -1 +1 @@
-export { Component as Heading } from './component';
```
davidpelayo: 😨

PR #45, opened 2018-04-03
TE-266 Button Additional props

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-266)

Add `hasShadow`, `icon`, `isRounded`, `isSecondary` and `size` props to the existing component.

#### Second Iteration (after PR comments)
buttonsv2

#### First Iteration
buttons


src/components/elements/Button/component.js
```diff
@@ -11,12 +13,26 @@ export const Component = ({
...
   <Button
     disabled={isDisabled}
     floated={isPositionedRight ? 'right' : 'left'}
     loading={isLoading}
+    secondary={isSecondary}
+    size={size}
+    className={cx({
+      'has-icon': !!icon,
+      'has-shadow': !!hasShadow,
+      'is-squared': !isRounded,
```
marcdavi-es: Perhaps we can simplify this by using the `circular` prop from Semantic UI button?

src/components/elements/Button/component.js
```diff
@@ -11,12 +13,26 @@ export const Component = ({
...
+  size,
 }) => (
   <Button
     disabled={isDisabled}
     floated={isPositionedRight ? 'right' : 'left'}
     loading={isLoading}
+    secondary={isSecondary}
+    size={size}
+    className={cx({
+      'has-icon': !!icon,
```
marcdavi-es: Do we need this? In `SearchBar` I just passed down an Icon as a child?

src/components/elements/Button/component.js
davidpelayo: Maybe I misunderstood the objective of `circular`? I thought, from the docs, it's intended to render circular button borders, which is slightly different from what we want to achieve.
We only have two modes:
- Squared buttons (pure rectangles).
- Rounded buttons (almost pure rectangles - minor rounded corners).

Whereas `circular` enforces aggressive round corners, and that's maybe something we do not want?

davidpelayo: It makes sense to use `circular`. I agree after thinking a bit more. Changing it.

src/components/elements/Button/Readme.md
```diff
@@ -20,3 +20,83 @@
   Submit
 </Button>
 ```
+
+### Variations
+
+#### Secondary
+
+```jsx
+<Grid style={{ backgroundColor: '#333' }}>
```
marcdavi-es: This example seems quite complicated. Do we need this whole Grid setup?

davidpelayo: I agree is not the best. Do you think we should go for a different component usage passing a button as child maybe? Or just a styled div to reflect this one needs a dark background layer?

src/components/elements/Button/component.js
```diff
@@ -11,12 +13,30 @@ export const Component = ({
...
+      'has-shadow': !!hasShadow,
+    })}
   >
+    {!icon ? null : (
+      <span className="icon">
+        {React.isValidElement(icon) ? <div className="icon">{icon}</div> : null}
+        {isString(icon) ? <Icon name={icon} /> : null}
+      </span>
+    )}
```
marcdavi-es: This icon thing seems complicated. Don't we just need 

`{icon && <Icon name={icon} />}`

?

Then we can limit the PropTypes to `string`. I think that's enough...?

src/components/elements/Button/Readme.md
marcdavi-es: For `Header` I just put it in a div...

screen shot 2018-04-03 at 17 56 22

PR #46, opened 2018-04-03
TE-269 Icon

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-269)

Creates component, examples and tests for Icon.

screen shot 2018-04-03 at 16 35 35


src/components/elements/Icon/component.js
```diff
@@ -0,0 +1,40 @@
...
+import { Paragraph } from 'typography/Paragraph';
+
+/**
+ * An icon is a glyph used to represent something else.
+ * @return {Object}
+ */
+export const Component = ({ isDisabled, label, name }) => (
+  <div>
+    <Icon
+      color={isDisabled ? 'grey' : undefined}
```
davidpelayo: Props should never setup color literals at a render level?
I know there is a enum with colors defined in Semantic, and that `inverted` controls how to contrast colors.
Maybe we should customize always via CSS and use mixins for contrast colors?

src/components/elements/Icon/component.js
```diff
@@ -0,0 +1,40 @@
...
+      inverted={isDisabled}
+      name={name}
+      size="large"
```
davidpelayo: This should exposed via props maybe, falling back to `large`?

src/components/elements/Icon/component.js
```diff
@@ -0,0 +1,40 @@
...
+import { Icon } from 'semantic-ui-react';
+
+import { Paragraph } from 'typography/Paragraph';
+
+/**
+ * An icon is a glyph used to represent something else.
+ * @return {Object}
+ */
+export const Component = ({ isDisabled, label, name }) => (
+  <div>
```
davidpelayo: Icons should be always render inline, I think. Using `<span>` maybe?

src/components/elements/Icon/component.js
```diff
@@ -0,0 +1,40 @@
...
+    <Icon
+      color={isDisabled ? 'grey' : undefined}
+      inverted={isDisabled}
+      name={name}
+      size="large"
+    />
+    {label && <Paragraph>{label}</Paragraph>}
```
davidpelayo: From the accessibility point of view, we should only use `<label>Insert icon here</label>` when a label is informed. Maybe we do not want to go there now?

src/components/elements/Icon/component.js
marcdavi-es: Here I'm just trying to use the existing API for Icon to get us where we need to be. This requires zero extra CSS.

The alternative is setting up a new custom class (`is-disabled` or whatever) and then writing extra CSS to get us to the same place.

I'm always in favour of less code...?

src/components/elements/Icon/component.js
marcdavi-es: I think for now I'm happy limiting this to one default. We can easily extend it later.

src/components/elements/Icon/component.js
davidpelayo: If we think we can live with that, totally agree. If it's a requirement to customize colors for inline content (including icons) - then we need it.
I haven't reviewed the Livingstone template layouts, but we have had cases were inline content (text and icons) needed to be restyled and coloured concretely given the user selected theme.

src/components/elements/Icon/component.js
davidpelayo: Agreed.

src/components/elements/Icon/component.js
marcdavi-es: Yeah for now I think we just want it hardcoded :)

src/components/elements/Icon/component.js
marcdavi-es: Changed

PR #49, opened 2018-04-04
TE-278 GoogleMap

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-278)

Creates component, examples and tests for the Map element.

### Any other notes
- Set up an intermediate component `ReactGoogleMap` to wrap the canonical implementation of `react-google-maps`
- Had to rename `Image` to `ResponsiveImage` to avoid conflict with the global `window.Image`, which the Google Maps API consumes. I'm not in love with `ResponsiveImage` and open to suggestions.


lib/react-google-maps/component.js
```diff
@@ -0,0 +1,39 @@
...
+  isShowingVagueMarker,
+  latitude,
+  longitude,
+}) => {
+  const coordinates = { lat: latitude, lng: longitude };
+  return (
+    <GoogleMap center={coordinates} options={mapOptions}>
+      {isShowingExactMarker && <Marker position={coordinates} />}
+      {isShowingVagueMarker && (
+        <Circle center={coordinates} options={circleOptions} />
```
davidpelayo: Better to resolve with a helper function which builds the object shape. In addition, we do not create scoped constants on each render (or update of the component itself).

```jsx
export const Component = ({
  isShowingExactMarker,
  isShowingVagueMarker,
  latitude,
  longitude,
}) => (
    <GoogleMap center={getCenterGivenLatLng(latitude, longitude)} options={mapOptions}>
    </GoogleMap>
);
```

lib/react-google-maps/component.js
```diff
@@ -0,0 +1,39 @@
...
+export const Component = ({
+  isShowingExactMarker,
+  isShowingVagueMarker,
+  latitude,
+  longitude,
+}) => {
+  const coordinates = { lat: latitude, lng: longitude };
+  return (
+    <GoogleMap center={coordinates} options={mapOptions}>
+      {isShowingExactMarker && <Marker position={coordinates} />}
```
davidpelayo: Force double coercion to boolean to avoid falsy values to render something different than `null`, like follows:
```jsx
{!!isShowingExactMarker && <Marker position={coordinates} />}
```

lib/react-google-maps/constants/circleOptions.js
```diff
@@ -0,0 +1,14 @@
+const circleColor = '#ea4336';
```
davidpelayo: This constant should be handled in the component/app context, not at lib level.

lib/react-google-maps/constants/mapOptions.js
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Predicated by Google Maps API v3
+ * See https://developers.google.com/maps/documentation/javascript/reference/3.exp/map#MapOptions
+ */
+export const mapOptions = {
+  mapTypeControl: false,
+  streetViewControl: false,
+  styles: mapStyles,
+  zoom: 13,
```
davidpelayo: Same here. This is something to be handled via component props too.

lib/react-google-maps/constants/circleOptions.js
```diff
@@ -0,0 +1,14 @@
...
+ * See https://developers.google.com/maps/documentation/javascript/reference/3.exp/polygon#CircleOptions
+ */
+export const circleOptions = {
+  clickable: false,
+  fillColor: circleColor,
+  fillOpacity: 0.5,
+  radius: 500, // in metres
+  strokeColor: circleColor,
+  strokeOpacity: 1,
+};
```
davidpelayo: These should be customizable via props too.

lib/react-google-maps/constants/mapStyles.js
```diff
@@ -0,0 +1,261 @@
+/**
+ * Generated using https://mapstyle.withgoogle.com/
+ */
+export const mapStyles = [
```
davidpelayo: This should be overridable via props, with fallback on this constant.

lib/react-google-maps/constants/reactGoogleMapsRequiredProps.js
```diff
@@ -0,0 +1,16 @@
...
+import { Card } from 'semantic-ui-react';
+
+/**
+ * Predicated by react-google-maps
+ * See https://tomchentw.github.io/react-google-maps/#usage--configuration
+ */
+export const reactGoogleMapsRequiredProps = {
+  // react-google-maps `withScriptjs` props
+  googleMapURL:
+    'https://maps.googleapis.com/maps/api/js?v=3.exp&libraries=geometry,drawing,places',
```
davidpelayo: An API key should be injectable to the widget, resulting on composing this URL with it.

lib/react-google-maps/container.js
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Canonical implementation for react-google-maps.
+ * See https://tomchentw.github.io/react-google-maps/#usage--configuration
+ */
+export const Container = compose(
+  withProps(reactGoogleMapsRequiredProps),
+  withScriptjs,
+  withGoogleMap
+)(Component);
```
davidpelayo: Brilliant ✨

lib/react-google-maps/component.js
```diff
@@ -0,0 +1,39 @@
...
+
+import { mapOptions, circleOptions } from './constants';
+
+/**
+ * Canonical implementation for react-google-maps.
+ * See https://tomchentw.github.io/react-google-maps/#usage--configuration
+ */
+export const Component = ({
+  isShowingExactMarker,
+  isShowingVagueMarker,
```
davidpelayo: I would rename this to something easier to understand, similar to the concept of: `isHidingExactLocation` (as Strava does) or `isShowingApproximateMarkerPosition` or similar.

lib/react-google-maps/constants/mapOptions.js
marcdavi-es: Same thoughts as here https://github.com/lodgify/lodgify-ui/pull/49#discussion_r179401020

lib/react-google-maps/constants/circleOptions.js
marcdavi-es: Same thoughts as here #49 (comment)

lib/react-google-maps/constants/mapStyles.js
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/49#discussion_r179401020

lib/react-google-maps/constants/reactGoogleMapsRequiredProps.js
marcdavi-es: I don't think we need a key?

lib/react-google-maps/component.js
marcdavi-es: Changed.

lib/react-google-maps/component.js
marcdavi-es: Changed.

lib/react-google-maps/component.js
marcdavi-es: Yeah `Vague` was a bullshit name :/

I've changed to `isShowingExactLocation` and `isShowingApproximateLocation`, which I think are better names.

I have avoided a ternary though i.e. `isHidingExactLocation ? <Circle /> : <Marker />` because I want to leave the consumer with three options:

- Show marker
- Show circle
- Show nothing

marcdavi-es:
> Why is there a problem on having <Image /> named as Image? Is it a consequence of the styleguide or what's the exact reason?

Si - because Styleguidist exposes all the components on the global object, `Image` conflicts with `window.Image`.

I discovered it now because Google Maps uses `window.Image` to create some part of the map.

PR #50, opened 2018-04-04
TE-273 KeyFacts

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-273)

Creates component, examples and tests for KeyFacts

### Any other notes
Made a couple of tweaks to existing styles to play nicely with KeyFacts.


src/components/widgets/KeyFacts/component.js
```diff
@@ -0,0 +1,53 @@
...
+import { GridColumn } from 'layout/GridColumn';
+import { Heading } from 'typography/Heading';
+import { IconCard } from 'elements/IconCard';
+
+/**
+ * The standard widget for displaying key facts about a property.
+ * @returns {Object}
+ */
+export const Component = ({ keyFacts, width }) => (
+  <GridColumn width={width}>
```
davidpelayo: The widget itself can't render a column, forcing and conditioning to be rendered inside a `<Grid>` system.

This grid column needs to be out of this scope or defining a whole Grid structure inside of the widget.

marcdavi-es: 🤔 In my mind, all widgets have to be wrapped in a `GridColumn`. Let's talk about this later/next week.

PR #58, opened 2018-04-10
TE-264 PropertyFooter

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-264)

Adding the `<PropertyFooter />` widget.

### Any other notes
- There is an `<Icon />` rendered in place of the `<PropertySummary />`, which is ongoing for now.
  - Property Summary info needs to be added by props and propagated down to the `<PropertySummary />` widget.
- Found certain icons don't render correctly. [TE-282 created for this](https://youtrack.lodgify.net/issue/TE-282).
- Found that Search buttons need a creation of a `<SearchButton />` component receiving props like `isMultiProperties` by props and caring about how different texts display. [TE-281 created for this](https://youtrack.lodgify.net/issue/TE-281).
  - If single property: Search with icon.
  - If multi properties: Check Availability without icon.
- Stickiness of this bar will be developed in a HoC apart.

screen shot 2018-04-10 at 12 00 41 pm

### DIFF
--- a/src/components/widgets/PropertyFooter/Readme.md
+++ b/src/components/widgets/PropertyFooter/Readme.md
@@ -1,0 +1,12 @@
+```jsx
+const { guestsOptions, locationOptions } = require('./mock-data/options');
+
+<Grid>
+  <GridRow>
+    <PropertyFooter
+      guestsOptions={guestsOptions}
+      locationOptions={locationOptions}
+    />
+  </GridRow>
+</Grid>
+```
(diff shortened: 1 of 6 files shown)


davidpelayo:
As commented on the [related YouTrack issue](https://youtrack.lodgify.net/issue/TE-264#comment=4-3768), we'll proceed to refactor this idea for a better and more common sense one, considering our current set of widgets in place.

PR #66, opened 2018-04-11
TE-312 Sprint eight retrospective - Demo DRYer tests.

marcdavi-es:
Create proposed expect helpers to check props and children.


davidpelayo:
@marcdavi-es btw, the PR doesn't have to do with sprinting, just with Youtrack issues so we keep track of changes individually. Creating a small youtrack task for this and linking them together as we do for the regular task would be awesome :rage4:

PR #80, opened 2018-04-16
TE-309 reviews widget

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-309)

### What things do this PR do?
- The Reviews Widget
- Minor upgrades on dependencies.


.eslintignore
```diff
@@ -4,6 +4,9 @@
 # Ignore coverage reports generated by Jest
 coverage/
 
+# Ignore gh-pages/
+gh-pages/
```
marcdavi-es: How come we need to do this?

davidpelayo: If you, for whatever reason (normally checking whether the built asset is OK in your local env) have the `gh-pages` folder created because you did run `npm run docs:build`, then linting and coverage take ages.

src/components/widgets/Reviews/component.js
```diff
@@ -0,0 +1,117 @@
...
+              reviewResponse={reviewResponse}
+              reviewerStayDate={reviewerStayDate}
+              reviewText={reviewText}
+              reviewTitle={reviewTitle}
+            />
+            <Divider />
+          </GridColumn>
+        </GridRow>
+      )
+    )}
```
marcdavi-es: Perhaps it's cleaner to not destructure and restructure the props?

```
{reviews.map(
  (review, index) => (
    <GridRow key={getUniqueKey(review.reviewText, index)}>
      <GridColumn width={12}>
        <Review {...review}/>
        <Divider />
      </GridColumn>
    </GridRow>
  )
)}
```

.eslintignore
marcdavi-es: OK :)

PR #81, opened 2018-04-17
TE-306 Payment Information widget

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-306)

### What one thing does this PR do?
- Creates the `PaymentInformation` widget.

### Any other notes
- `Statistic` semantic-ui view is used.

paymentinfo

### DIFF
--- a/src/components/typography/Paragraph/component.js
+++ b/src/components/typography/Paragraph/component.js
@@ -8,24 +8,14 @@ const TINY = 'tiny';
  * A paragraph provides text content
  * @return {Object}
  */
-export const Component = ({ children, size, isWord, isBoldWord }) => {
-  if (isBoldWord) {
-    return <b className={size === TINY ? size : ''}>{children}</b>;
-  }
-
-  if (isWord) {
-    return <span className={size === TINY ? size : ''}>{children}</span>;
-  }
-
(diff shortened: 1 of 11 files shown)


src/components/typography/Paragraph/component.js
```diff
@@ -8,19 +8,33 @@ const TINY = 'tiny';
...
+  if (isBoldWord) {
+    return <b className={size === TINY ? size : ''}>{children}</b>;
+  }
+
+  if (isWord) {
+    return <span className={size === TINY ? size : ''}>{children}</span>;
+  }
+
+  return <p className={size === TINY ? size : ''}>{children}</p>;
+};
```
marcdavi-es: I'm not sure about extending the `Paragraph` API for these cases.

The term 'paragraph' has semantic importance in both design and html. It's specifically _not_ a single word.

I would prefer to change `Paragraph`'s `props.children` so that it can accept markup including `<b>` and `<span>` (and any other) tags for these use cases.

What do you think?

marcdavi-es: What about...

```
  <Paragraph size="medium">
     <b>{taxesText}</b>{taxesDescriptionText}
  </Paragraph>
```

PR #87, opened 2018-04-19
TE-301 Change button

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-301)

Makes consequent changes to Button element component and styles.


src/semantic/src/themes/livingstone/elements/button.overrides
```diff
@@ -24,6 +24,9 @@
 /* Box Shadow */
 
 /* Icon */
+.ui.button > .icon:not(.button) > svg {
```
davidpelayo: `:not(.button)`?

PR #103, opened 2018-04-20
TE-303 Subheading

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-303)

Implements a new `Subheading` component.

screen shot 2018-04-20 at 19 01 11


src/components/typography/Subheading/component.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * A subheading provides concise information about a block of content.
+ * @return {Object}
+ */
+export const Component = ({ children }) => (
+  <h5 className="ui sub header">{children}</h5>
```
davidpelayo: Not a big deal but this shouldn't be a `h5` element. Please change to a regular `p` or `span`.
Heading elements ideally need to be used sequentially, and since this is a component to be used anywhere, it's better to just replace it by a more general tag name.

PR #106, opened 2018-04-23
TE-334 PaymentInformation responsive layout

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-334)

Gives the PaymentInformation widget a mobile friendly layout

### Any other notes
image

### DIFF
--- a/src/components/widgets/PaymentInformation/component.js
+++ b/src/components/widgets/PaymentInformation/component.js
@@ -26,7 +26,7 @@ export const Component = ({
   notesText,
   extraNotesText,
 }) => (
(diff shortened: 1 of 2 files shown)


src/components/widgets/PaymentInformation/component.js
```diff
@@ -26,7 +26,7 @@ export const Component = ({
   notesText,
   extraNotesText,
 }) => (
-  <Grid>
+  <Grid stackable>
```
marcdavi-es: Can we simplify this?

PR #108, opened 2018-04-23
TE-322 Form responsive

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-322)

Adds responsive styles for Form collection.

### Any other notes
Tweaks examples which consume `Form`

screen shot 2018-04-23 at 13 33 10

### DIFF
--- a/src/semantic/src/themes/livingstone/collections/form.overrides
+++ b/src/semantic/src/themes/livingstone/collections/form.overrides
@@ -19,9 +19,13 @@
   /* Form Label */
 
   /* Input */
-  .field .ui.input,
-  .fields .field .ui.input {
-    max-width: unset;
+  .field,
+  .fields .field {
+
+    .ui.input,
+    .ui.dropdown {
+      max-width: 100%;
+    }
   }
 
   /* Select */
@@ -78,6 +82,13 @@
 --------------------*/
 
   /* Grouped Fields */
+  &[class*="equal width"] .fields > .field,
+  [class*="equal width"].fields > .field {
+    // This prevents any field from extending beyond the right edge of its
+    // containing field group. Flex sizing of each field is delegated to
+    // `semantic/src/definitions/collections/form.less` line 950
+    width: 0%;
+  }
 
   /* Inline */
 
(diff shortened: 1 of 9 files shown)


src/semantic/src/themes/livingstone/collections/form.overrides
```diff
@@ -78,6 +82,13 @@
 --------------------*/
 
   /* Grouped Fields */
+  &[class*="equal width"] .fields > .field,
+  [class*="equal width"].fields > .field {
+    // This prevents any field from extending beyond the right edge of its
+    // containing field group. Flex sizing of each field is delegated to
+    // `semantic/src/definitions/collections/form.less` line 950
+    width: 0%;
```
davidpelayo: ? I don't get it. Why `width: 0%`?

src/semantic/src/themes/livingstone/globals/site.variables
```diff
@@ -133,6 +133,14 @@
      Breakpoints
 --------------------*/
 
+@mobileBreakpoint: 1px;
+@tabletBreakpoint: 600px;
+@computerBreakpoint: 1000px;
+@largeMonitorBreakpoint: 1920px;
+@widescreenMonitorBreakpoint: 1920px;
+
+@mobileScreen: ~"only screen and (min-width: @{mobileBreakpoint}) and (max-width: @{largestMobileScreen})";
```
davidpelayo: I believe this is not right?
Mobile breakpoint usually stands for the breakpoint in which lower values defines the mobile layout.
Why then the need of declaring a min-width of 1px?

Jonnygwi: +1

src/semantic/src/themes/livingstone/collections/form.overrides
marcdavi-es: It allows the children of an `equal width` field group to play nicely with `flex`. They grow to fill the available space.

marcdavi-es: Sadly no. I know this seems hacky but it's the most reliable thing we can do.

PR #109, opened 2018-04-23
TE-328 Responsive keyfacts

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-328)

Gives the `KeyFacts` widget a mobile friendly layout

screen shot 2018-04-23 at 6 06 15 pm
screen shot 2018-04-23 at 6 06 22 pm
screen shot 2018-04-23 at 6 06 31 pm

### DIFF
--- a/src/components/widgets/KeyFacts/component.js
+++ b/src/components/widgets/KeyFacts/component.js
@@ -4,6 +4,8 @@ import { Label } from 'semantic-ui-react';
 
 import { getUniqueKey } from 'lib/get-unique-key';
 import { Heading } from 'typography/Heading';
+import { Grid } from 'layout/Grid';
+import { GridColumn } from 'layout/GridColumn';
 import { IconCard } from 'elements/IconCard';
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/widgets/KeyFacts/component.js
```diff
@@ -4,27 +4,42 @@ import { Label } from 'semantic-ui-react';
...
-          isDisabled={isDisabled}
-          isFilled
-          key={getUniqueKey(label, index)}
-          label={label}
-          name={iconName}
-        />
-      ))}
-    </Label.Group>
-  </div>
+  <Grid>
```
marcdavi-es: Is the `Grid` necessary?

src/components/widgets/KeyFacts/component.js
```diff
@@ -4,27 +4,42 @@ import { Label } from 'semantic-ui-react';
...
+    <GridColumn width={12}>
+      <Heading>Key facts</Heading>
+    </GridColumn>
+    <GridColumn width={12}>
+      <Label.Group>
```
marcdavi-es: I don't think we should use `Label.Group` AND `Grid` - one or the other will do.

marcdavi-es:
A few things:

1. A couple comments inline. I'm not sure we need the two extra `Grids`
2. [The issue card](https://youtrack.lodgify.net/issue/TE-328) includes a note about making sure `IconCard` displays with consistent height and width. Let's include that?
3. Let's take the opportunity to fix the example which is displaying weirdly because of the legacy wrapping `GridRow`. The Readme.md can just be...

````
```jsx
const { keyFacts } = require('./mock-data/keyFacts');

<KeyFacts keyFacts={keyFacts} />
```

### Variations

#### Width

```jsx
const { keyFacts } = require('./mock-data/keyFacts');

<div>
  <KeyFacts keyFacts={keyFacts} />
  <Divider hasLine />
  <KeyFacts keyFacts={keyFacts} width={6} />
</div>
```
````

which gives us...

screen shot 2018-04-23 at 17 15 00

src/components/widgets/KeyFacts/component.js
davidpelayo: It's actually not needed. But I do read better the code having semantically a `Grid` component in mind rather than a plain and speechless `div` 🌹

PR #114, opened 2018-04-24
TE-283 Static Sticky HoC

davidpelayo:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-283)

Adding a `StaticSticky` HoC to be used in `PropertyFooter` as a basic concept - not dynamic scrolling event handling

### Any other notes
- Upgraded the npm dependencies.
- Tried [react-sticky](https://github.com/captivationsoftware/react-sticky), `display: sticky` and my own custom implementation. Then I realized we don't need dynamic stickiness behaviour for now.
- Added a question to [here](https://github.com/styleguidist/react-styleguidist/issues/953) wondering what's the best way to document sticky components.

hoc


lib/react-static-sticky/component.js
```diff
@@ -0,0 +1,57 @@
...
+  children: null,
+  isRelativeSticky: false,
+  willStickToBottom: false,
+  styles: {},
+};
+
+Component.propTypes = {
+  /** The children to render. */
+  children: PropTypes.node,
+  /** Is the children going to be render in relative sticky mode. */
```
Jonnygwi: This prop type could do with more description. For example what relative sticky mode means

lib/react-static-sticky/component.js
```diff
@@ -0,0 +1,57 @@
...
+      display: 'inline-block',
+      height: '100%',
+      width: '100%',
+      ...styles,
+    }}
+  >
+    <div
+      style={{
+        position: !isRelativeSticky ? 'fixed' : 'absolute',
+        bottom: willStickToBottom ? 0 : 'initial',
```
Jonnygwi: maybe rename prop to `isStuckToBottom` to be consistent with `isRelativeSticky`

lib/react-static-sticky/component.js
```diff
@@ -0,0 +1,57 @@
...
+  styles: {},
+};
+
+Component.propTypes = {
+  /** The children to render. */
+  children: PropTypes.node,
+  /** Is the children going to be render in relative sticky mode. */
+  isRelativeSticky: PropTypes.bool,
+  /** Will the children stick to bottom. */
+  willStickToBottom: PropTypes.bool,
```
Jonnygwi: 👆

lib/react-static-sticky/component.js
```diff
@@ -0,0 +1,57 @@
...
+    </div>
+  </div>
+);
+
+Component.displayName = 'StaticSticky';
+
+Component.defaultProps = {
+  children: null,
+  isRelativeSticky: false,
+  willStickToBottom: false,
```
Jonnygwi: See my previous comment on `isStuckToBottom`

davidpelayo:
@marcdavi-es I didn't go for existing solutions because they all get introduced into the subscribing and dispatching beautiful world of events, while scrolling. Which I think we need to avoid for now, until it gets needed due some dynamic stickiness behaviour as requirement.

Also, in our case, we probably need things sticky relative to parent containers, not the viewport itself. So that we avoid having fixed elements to the viewport.

If we want to have a classical sticky implementation, then yes, let's use the Semantic UI React implementation for absolute stickiness to viewport.

marcdavi-es:
@davidpelayo Closing this for now. Let me know what you want to do with it :)

PR #127, opened 2018-04-27
TE-351 Add placeholder screen for mobile view in production

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-351)

Adds placeholder screen for mobile view in production

screen shot 2018-04-27 at 17 57 25

### DIFF
--- a/styleguide/styleguide-components/MobilePlaceholder/component.js
+++ b/styleguide/styleguide-components/MobilePlaceholder/component.js
@@ -1,0 +1,95 @@
+import React from 'react';
+import Styled from 'rsg-components/Styled';
+
+import LogoRenderer from '../LogoRenderer';
+
+import { getMailToHref } from './getMailToHref';
+
+const styles = ({ borderRadius, color, mq, space }) => ({
+  screen: {
+    background: color.baseBackground,
+    display: 'none',
+    height: '100vh',
+    overflowY: 'scroll',
+    padding: space[4],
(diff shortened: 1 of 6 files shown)


styleguide/styleguide-components/MobilePlaceholder/component.js
```diff
@@ -0,0 +1,95 @@
...
+      isolate: false,
+      color: color.white,
+      textDecoration: 'none',
+    }
+  }
+})
+
+class Markup extends React.Component {
+
+  state = {
```
Jonnygwi: Do you need to do this if you set a defaultValue to the input?

marcdavi-es: I prefer to be explicit about the state which we intend the component to hold.

davidpelayo: Totally agree with @marcdavi-es

styleguide/styleguide-components/StyleGuideRenderer/getIsDevelopmentServer.js
```diff
@@ -0,0 +1 @@
+export const getIsDevelopmentServer = () => !!webpackHotUpdate;
```
davidpelayo: I would prefer to inject environment variables (via webpack) instead, but this should work fine.

PR #131, opened 2018-05-01
TE-350 Side-by-side editor

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-350)

Creates a side-by-side code editor panel on wide screens.

### Any other notes
- on screens less than 1200px, it falls back on the previous editor setup.
- I am not as religious in the organisation of styleguidist components as I am with our own. I hope you don't tell me off too much.

**Hide/Show***
widescreen1

**Fallback on smaller screens***
normalscreen

### DIFF
--- a/styleguide/styleguide-components/PlaygroundRenderer/component.js
+++ b/styleguide/styleguide-components/PlaygroundRenderer/component.js
@@ -1,0 +1,131 @@
+/**
+ *
+ * Lodgify UI overwrite of
+ * React Styleguidist PlaygroundRenderer.
+ *
+ * Most of this file is a duplicate of
+ * https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Playground/PlaygroundRenderer.js
+ *
+ * Search 'Lodgify UI' for custom script, styles and markup.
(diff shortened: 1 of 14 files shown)


styleguide/styleguide-components/PlaygroundRenderer/component.js
```diff
@@ -0,0 +1,131 @@
...
+// Lodgify UI import start
+import { compose } from 'recompose';
+
+import { withResponsive } from 'lib/with-responsive';
+// Lodgify UI import end
+
+export const styles = ({ space, color, borderRadius, mq }) => ({
+	root: {
+		marginBottom: space[4],
+  	// Lodgify UI styles start
```
davidpelayo: Indentation looks weird

marcdavi-es: Yeah, it is weird.

For the components duplicated from styleguidist, I've chosen to go with their indentation style. It makes it easier to come back and compare our custom implementation later.

I think it's the lesser of two evils. Can you live with that?

PR #134, opened 2018-05-02
TE-357 Alpha ribbin z-index fix

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-357)

It puts the alpha ribbon behind the mobile placeholder screen but still in front of the code blocks.

### Any other notes
image

image

### DIFF
--- a/styleguide/styleguide-components/MobilePlaceholder/component.js
+++ b/styleguide/styleguide-components/MobilePlaceholder/component.js
@@ -15,7 +15,7 @@ const styles = ({ borderRadius, color, mq, space }) => ({
     position: 'fixed',
     visibility: 'hidden',
     width: '100vw',
-    zIndex: 2,
+    zIndex: 11,
 
     [mq.small]: {
       display: 'block',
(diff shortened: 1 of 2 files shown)


styleguide/styleguide-components/MobilePlaceholder/component.js
```diff
@@ -15,7 +15,7 @@ const styles = ({ borderRadius, color, mq, space }) => ({
     position: 'fixed',
     visibility: 'hidden',
     width: '100vw',
-    zIndex: 2,
+    zIndex: 11,
```
davidpelayo: We need to stablish a hashmap for better zIndex management, to avoid this dancing.
Not maybe for this purpose but yes for the templates themselves.
In SASS it's usually something like this:
```sass
$z-layers : (
  "modal": 100,
  "overlay": 10,
  "default": 1,
  "below": -1
);

@function z($layer) {
  @if not map-has-key($z-layers, $layer) {
    @warn "No layer found for `#{$layer}` in $z-layers map. Property omitted.";
  }

  @return map-get($z-layers, $layer);
}
```

I guess you guys capture the idea, right? cc/ @marcdavi-es @Jonnygwi

marcdavi-es: Si. Have created https://youtrack.lodgify.net/issue/TE-369

PR #140, opened 2018-05-07
TE-362 Propose package api

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-362)

Adds the API for all the components to be consumed

### Any other notes
Added documentation for importing the styles and the components

image


davidpelayo:
I would add specifically a description on why the styles should be imported.

PR #148, opened 2018-05-09
TE-362 Fix component imports

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-362)

Updates the imports in the documentation 

### Any other notes
Also fixed an uppercasing error in the Getting Started section

image


styleguide/config/options/componentPathLine.js
```diff
@@ -0,0 +1,14 @@
+const path = require('path');
+
+const getLastFolderFromFilePath = filePath => {
+  const directory = path.dirname(filePath);
+  return directory.substring(directory.lastIndexOf('/') + 1);
```
marcdavi-es: Think we should use `path.sep` to be OS-neutral.

PR #164, opened 2018-05-28
TE-398 Add Table collection

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-398)

Exposes a `Table` collection from Lodgify UI

### Any other notes
kapture 2018-05-30 at 15 37 19

### DIFF
--- a/src/components/collections/Table/Readme.md
+++ b/src/components/collections/Table/Readme.md
@@ -1,0 +1,47 @@
+```jsx
+const headers = [
+  'Rates',
+  'Mon',
+  'Tue',
+  'Wed',
+  'Thu',
+  'Fri',
+  'Sat',
+  'Sun'
+];
+
+const body = [
+  ['Default rate', '40', '30', '40', '40', '50', '60', '70'],
+  ['Mid Season rate', '40', '40', '50', '60', '60', '70', '80'],
+  ['High Season rate', '45', '45', '55', '65', '65', '75', '85'],
+];
+
+<Table tableHeadings={headers} tableBody={body} />
+```
+
+### Content
+
+React components can be used as table items
+
+```jsx
+const { dropdownOptions } = require('./mock-data/props');
+
+const headers = [
+  <Dropdown key="location-dropdown" options={dropdownOptions} />,
+  'Mon',
+  'Tue',
+  'Wed',
+  'Thu',
+  'Fri',
+  'Sat',
+  'Sun'
+];
+
+const body = [
+  ['Default rate', '40', '30', '40', '40', '50', '60', '70'],
+  ['Mid Season rate', '40', '40', '50', '60', '60', '70', '80'],
+  ['High Season rate', '45', '45', '55', '65', '65', '75', '85'],
+];
+
+<Table tableHeadings={headers} tableBody={body} tableId="table-with-component"/>
+```
--- a/src/components/collections/Table/component.js
+++ b/src/components/collections/Table/component.js
@@ -1,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Table } from 'semantic-ui-react';
+
+import { buildKeyFromStrings } from 'utils/build-key-from-strings';
+
+/**
+ * Table is the Lodgify UI interface for the Semantic UI Table.
+ * @returns {Object}
+ */
+export const Component = ({ tableHeadings, tableBody, tableId }) => (
+  <Table basic="very" padded striped>
+    <Table.Header>
+      <Table.Row>
+        {tableHeadings.map((heading, index) => (
+          <Table.HeaderCell
+            key={buildKeyFromStrings(tableId, `heading${index}`)}
+          >
+            {heading}
+          </Table.HeaderCell>
+        ))}
+      </Table.Row>
+    </Table.Header>
+    <Table.Body>
+      {tableBody.map((row, rowIndex) => (
+        <Table.Row key={buildKeyFromStrings(tableId, `row${rowIndex}`)}>
+          {row.map((cell, cellIndex) => (
+            <Table.Cell
+              key={buildKeyFromStrings(
+                tableId,
+                `cell${cellIndex}`,
+                `row${rowIndex}`
+              )}
+            >
+              {cell}
+            </Table.Cell>
+          ))}
+        </Table.Row>
+      ))}
+    </Table.Body>
+  </Table>
+);
+
+Component.defaultProps = {
+  tableId: '',
+};
+
+Component.displayName = 'Table';
+
+Component.propTypes = {
+  /** The column heading represented in the table */
+  tableHeadings: PropTypes.arrayOf(PropTypes.node).isRequired,
+  /** The data making up the body of the table */
+  tableBody: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.node)).isRequired,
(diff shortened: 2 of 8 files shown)


src/components/collections/Table/Readme.md
```diff
@@ -0,0 +1,20 @@
...
+];
+
+const body = [
+  ['One', 30, 30, 40, 40, 50, 60, 70],
+  ['Two', 40, 40, 50, 60, 60, 70, 80],
+  ['Three', 45, 45, 55, 65, 65, 75, 85],
+];
+
+<Table tableHeaders={headers} tableBody={body} />
+```
```
marcdavi-es: Perhaps we ought to include an example (Under a `Content` heading?) which shows the `Table` dealing with React elements as input?

src/components/collections/Table/component.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Table } from 'semantic-ui-react';
+
+import { getUniqueKey } from 'utils/get-unique-key';
+
+/**
+ * Table is the Lodgify UI interface for the
+ * Semantic UI Table.
```
marcdavi-es: I don't think this needs to be on a new line?

src/components/collections/Table/component.js
```diff
@@ -0,0 +1,44 @@
...
+ * @returns {Object}
+ */
+export const Component = ({ tableHeaders, tableBody }) => (
```
marcdavi-es: I think this should be `tableHeadings` for consistency.

In Lodgify UI, `headings` are emphasised strings, `header` is the thing at the top of the whole page.

src/components/collections/Table/component.js
```diff
@@ -0,0 +1,44 @@
...
+        {tableHeaders.map((header, index) => (
+          <Table.HeaderCell key={getUniqueKey(header, index)}>
+            {header}
+          </Table.HeaderCell>
+        ))}
+      </Table.Row>
+    </Table.Header>
+    <Table.Body>
+      {tableBody.map((row, index) => (
+        <Table.Row key={getUniqueKey('tableRow', index)}>
```
marcdavi-es: I'm not sure this will give us unique keys. What if we have more than one table in the page? Try using some string from the row object?

src/components/collections/Table/component.js
```diff
@@ -0,0 +1,44 @@
...
+          {row.map((cell, index) => (
+            <Table.Cell key={getUniqueKey('tableCell', index)}>
```
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/164/files#r191366350

src/components/collections/Table/component.spec.js
```diff
@@ -0,0 +1,91 @@
...
+            .find(SemanticUITable.Row)
+            .first();
+          expectComponentToHaveChildren(
+            wrapper,
+            ...getArrayOfLengthOfItem(2, SemanticUITable.HeaderCell)
+          );
+        });
+      });
+
+      describe('the `SemanticUITable.Body` component', () => {
```
marcdavi-es: I don't think this `describe` should be nested in the one that starts on [line 46](https://github.com/lodgify/lodgify-ui/pull/164/files#diff-cf98ffa38bfdedd8caa2116731afb045R46)?

src/components/collections/Table/component.spec.js
```diff
@@ -0,0 +1,91 @@
...
+            .find(SemanticUITable.Row)
+            .at(1);
+          expectComponentToHaveChildren(
+            wrapper,
+            ...getArrayOfLengthOfItem(2, SemanticUITable.Cell)
+          );
+        });
+      });
+    });
+  });
```
marcdavi-es: Missing test cases for the contents of the `HeaderCell`s and `Cell`s

src/styles/semantic/themes/livingstone/collections/table.overrides
```diff
@@ -0,0 +1,86 @@
...
+/* Basic */
+
+/* Very Basic */
+.ui[class*="very basic"].table tbody tr {
+  border: none;
+}
+
+.ui[class*="very basic"].table tr td {
+  border: none;
+}
```
marcdavi-es: I think you can lose the borders by using the variable `@basicTableCellBorder`

src/styles/semantic/themes/livingstone/collections/table.overrides
```diff
@@ -0,0 +1,86 @@
...
+
+.ui[class*="very basic"].striped.table tbody tr:nth-child(1n) {
+  background-color: @greyOne;
+}
+// Respective style in semantic-ui has !important flag
+.ui[class*="very basic"].striped.table tbody tr:nth-child(2n) {
+  background-color: @pureWhite !important;
+}
```
marcdavi-es: I think you can do the background colors using variables:

`@basicTableCellBackground`
`@basicTableStripedBackground`

If that's true, we can lose this overrides files and use only a variables file, which is more betterer.

src/components/collections/Table/utils/getKeyForTableContent.js
```diff
@@ -0,0 +1,13 @@
...
+ * @param  {String} tableId
+ * @param  {(Object|String)} tableContent
+ * @param  {...Number} indexes
+ * @return {String}
+ */
+export const getKeyForTableContent = (tableId, tableContent, ...indexes) =>
+  buildKeyFromStrings(tableId, getKeyFromContent(tableContent), ...indexes);
+
+const getKeyFromContent = content =>
+  typeof content === 'object' ? content.key : content;
```
marcdavi-es: This whole `getKeyForTableContent` thing is now really complicated.

Now that we take an id in for table, can't we just use `buildKeyFromStrings` directly in the component like...

`key={buildKeyFromStrings(tableId, `row${rowIndex}`, `cell${cellIndex}`)}`

or something?

src/components/collections/Table/component.js
```diff
@@ -0,0 +1,53 @@
...
+  tableId: '',
+};
+
+Component.displayName = 'Table';
+
+Component.propTypes = {
+  /** The column heading represented in the table */
+  tableHeadings: PropTypes.arrayOf(PropTypes.node).isRequired,
+  /** The data making up the body of the table */
+  tableBody: PropTypes.arrayOf(PropTypes.node).isRequired,
```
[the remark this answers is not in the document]
Jonnygwi: It wasn't raising a warning but you are right.. Strange

src/components/collections/Table/component.js
```diff
@@ -0,0 +1,57 @@
...
+            </Table.Cell>
+          ))}
+        </Table.Row>
+      ))}
+    </Table.Body>
+  </Table>
+);
+
+Component.defaultProps = {
+  tableId: '',
```
davidpelayo: The default should be a pseudo-aleatory generated ID, instead of empty.
If `tableId` isn't required, we could have 2 different tables without `tableId` informed, having potential issues.

Jonnygwi: @davidpelayo as discussed with @marcdavi-es React will throw a warning when it comes across the duplicate keys. The alternative is to be more explicit about why/when the key is needed in the documentation. What do you think?

PR #169, opened 2018-06-04
TE-406 Table text alignment

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-406)

Aligns all table cells in the first column to the left and all others to the center

### Any other notes
image

### DIFF
--- a/src/components/collections/Table/component.js
+++ b/src/components/collections/Table/component.js
@@ -4,6 +4,8 @@ import { Table } from 'semantic-ui-react';
 
 import { buildKeyFromStrings } from 'utils/build-key-from-strings';
 
+import { getTextAlign } from './utils/getTextAlign';
+
 /**
  * Table is the Lodgify UI interface for the Semantic UI Table.
  * @returns {Object}
@@ -14,6 +16,7 @@ export const Component = ({ tableHeadings, tableBody, tableId }) => (
       <Table.Row>
         {tableHeadings.map((heading, index) => (
           <Table.HeaderCell
+            textAlign={getTextAlign(index)}
             key={buildKeyFromStrings(tableId, `heading${index}`)}
           >
             {heading}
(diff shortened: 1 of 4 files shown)


src/components/collections/Table/component.js
```diff
@@ -26,6 +27,7 @@ export const Component = ({ tableHeadings, tableBody, tableId }) => (
         <Table.Row key={buildKeyFromStrings(tableId, `row${rowIndex}`)}>
           {row.map((cell, cellIndex) => (
             <Table.Cell
+              textAlign={cellIndex === 0 ? 'left' : 'center'}
```
marcdavi-es: DRY this up by isolating as a util

`export const getTextAlign = index => index === 0 ? 'left' : 'center'`

src/components/collections/Table/utils/getTextAlign.spec.js
```diff
@@ -0,0 +1,17 @@
+import { getTextAlign } from './getTextAlign';
+
+describe('getTextAlign', () => {
+  const array = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
```
marcdavi-es: Do you think we should expand this to all numbers?

marcdavi-es: 🚨JOKE ALERT 🚨

src/components/collections/Table/component.js
```diff
@@ -26,6 +29,7 @@ export const Component = ({ tableHeadings, tableBody, tableId }) => (
         <Table.Row key={buildKeyFromStrings(tableId, `row${rowIndex}`)}>
           {row.map((cell, cellIndex) => (
             <Table.Cell
+              textAlign={getTextAlign(cellIndex)}
```
davidpelayo: It feels to me that the table component needs hovering/focus effects on rows and cells to navigate the content properly

src/components/collections/Table/utils/getTextAlign.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * @param  {Number} index
+ * @return {String}
+ */
+export const getTextAlign = index => (index === 0 ? 'left' : 'center');
```
davidpelayo: I would expect by this util to be able to give me any position, whereas is just a two positions conditioner. I think renaming is needed here, although it's just a detail

src/components/collections/Table/component.js
jams676: The table is not interactive at all though. There's a small off-white/gray color change between rows that should be enough to differentiate from rows. Let me know if you have any other suggestions examples, but I think hover is unnecessary

PR #184, opened 2018-06-12
TE-441 FeaturedProperty unit test refactor

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-441)

Refactors component and unit test to meet the same criteria of FeaturedRoomType component

### Any other notes
It also adds a the `getPropertyDescription` util and its unit test. 

screen shot 2018-06-12 at 16 02 21


src/components/general-widgets/FeaturedProperty/component.js
```diff
@@ -50,17 +52,13 @@ export const Component = ({
...
-};
-
 Component.propTypes = {
   /** The number of available bedrooms at the property. */
   bedroomsNumber: PropTypes.number.isRequired,
   /** The number of guests the property can accommodate. */
   guestsNumber: PropTypes.number.isRequired,
   /** The alternative text for the image to display. */
-  imageAlternativeText: PropTypes.string,
+  imageAlternativeText: PropTypes.string.isRequired,
```
marcdavi-es: I don't think we want this to be `.isRequired`

src/components/general-widgets/FeaturedProperty/component.spec.js
```diff
@@ -1,17 +1,24 @@
...
 import { Heading } from 'typography/Heading';
 
 import { Component as FeaturedProperty } from './component';
+import { getPropertyDescription } from './utils/getPropertyDescription';
 
 const props = {
   bedroomsNumber: 3,
-  imageUrl: '🐱🐱',
   guestsNumber: 3,
+  imageAlternativeText: 'this is a tag',
```
marcdavi-es: Not required.

src/components/general-widgets/FeaturedProperty/component.spec.js
```diff
@@ -105,91 +91,78 @@ describe('<FeaturedProperty />', () => {
...
-          disabled: true,
-          maxRating: 5,
-          rating: Math.round(props.ratingNumber),
-          size: 'tiny',
-        })
-      );
+      const wrapper = getFeaturedProperty()
+        .find(Card.Description)
+        .at(2)
+        .find(Rating);
```
marcdavi-es: `const wrapper = getFeaturedProperty().find(Rating);`

is fine because we already know `Rating` is in the right place.

marcdavi-es:
Pretty good. A couple of things inline.

Please reorganise the commits like...

```
refactor(FeaturedProperty): isolate getPropertyDescription util
refactor(FeaturedProperty): consume @lodgify/enzyme-jest-expect-helpers in unit tests
refactor(FeaturedProperty): remove the mock-data folder
```

Jonnygwi:
Not a big one but PR name is miss leading. `TE-441 enzyme jest expect helpers refactor` sounds like you are refactoring the helpers not implementing the helpers into the already existing tests. Change PR name to `TE-441 FeaturedProperty unit test refactor`

src/components/general-widgets/FeaturedProperty/Readme.md
```diff
@@ -1,10 +1,8 @@
 ```jsx
-const { twoCats } = require('./mock-data/images');
```
davidpelayo: Nice!

src/components/general-widgets/FeaturedProperty/utils/getPropertyDescription.js
```diff
@@ -0,0 +1,7 @@
+/**
+ *  @param {Number} guestsNumber
+ *  @param {Number} bedroomsNumber
+ *  @returns {String}
+ */
+export const getPropertyDescription = (guestsNumber, bedroomsNumber) =>
+  `Guests: ${guestsNumber} | Bedrooms: ${bedroomsNumber}`;
```
davidpelayo: Good that we isolate the literals resolution since we will have to translate them using the i18n data.

src/components/general-widgets/FeaturedProperty/component.spec.js
```diff
@@ -1,17 +1,24 @@
...
 import { Heading } from 'typography/Heading';
 
 import { Component as FeaturedProperty } from './component';
+import { getPropertyDescription } from './utils/getPropertyDescription';
 
 const props = {
   bedroomsNumber: 3,
-  imageUrl: '🐱🐱',
   guestsNumber: 3,
+  imageAlternativeText: '',
```
davidpelayo: The alternative text should be required within the component, no default to `' '` should be done.

marcdavi-es: @davidpelayo we don't want the `imageAlternativeText` to be required - it's not always available in the model.

This line is unnecessary though - there's no need to have this prop included in the test props.

davidpelayo: Let's change the model then.

red-runner: @marcdavi-es and @davidpelayo I also think that alt is a very important asset. Yesterday I spoke with Marc and he claimed, if I understood correctly, that *alt* will not always be provided and therefore we should not make it mandatory; unless we want alt to be a random string.  

That being said, I switched the test according to Marc's observations. However, let me know the final decision as I'll be happy to switch the unit test again.

marcdavi-es: We're confusing two issues here

1. **Web content policy**: The web works better if images have altText. We are all agreed on that.
2. **Component functionality**. The Image component can work without altText.

`PropTypes.*.isRequired` is for enforcing 2., not 1. 

Even if we change the shape of our model, we can't guarantee...

- that there is any data there in the case of preexisting websites
- that future users will bother to enter alt text data when they make new websites
- that other consumers of lodgify-ui will have the same opinion of web policy as we do. 

In all of these cases, `Image` needs to (and can) still work.

So... no `.isRequired` here

Jonnygwi:
👍

src/components/general-widgets/FeaturedProperty/component.spec.js
marcdavi-es: @davidpelayo I'm not sure I agree that a guessed value derived from available strings is better than an empty value?

davidpelayo: They just mean different things.
The spec describes it clearly: https://www.w3.org/TR/2011/WD-html5-author-20110705/the-img-element.html#alt

marcdavi-es: Ah. I didn't know that there was a general imperative for alt text. Thanks.

BUT, from the same spec...

```
### Images whose contents are not known

In some unfortunate cases, there might be no alternative text available at all, either because the image is obtained in some automated fashion without any associated alternative text (e.g. a Webcam), or because the page is being generated by a script using user-provided images where the user did not provide suitable or usable alternative text (e.g. photograph sharing sites), or because the author does not himself know what the images represent (e.g. a blind photographer sharing an image on his blog).

In such cases, the alt attribute may be omitted, but one of the following conditions must be met as well:

- The title attribute is present and has a non-empty value.
- The img element is in a figure element that contains a figcaption element that contains content other than inter-element whitespace, and, ignoring the figcaption element and its descendants, the figure element has no text node descendants other than inter-element whitespace, and no embedded content descendant other than the img element.

```

So I think it's OK to omit. We're going to try to populate the title in most cases.

Also - we have to find a compromise between the official spec and general practice. Looking at e.g. AirBnb for a closely matching use case which has scale, they don't use `alt` or `title` at all...

screen shot 2018-06-15 at 10 39 08

(from https://www.airbnb.com/rooms/833602?)

PR #186, opened 2018-06-12
TE-446 Remove margin-top from .ui.modal > .ui.card

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-446)

Remove margin-top from .ui.modal > .ui.card

### DIFF
--- a/src/styles/semantic/themes/livingstone/modules/modal.overrides
+++ b/src/styles/semantic/themes/livingstone/modules/modal.overrides
@@ -23,6 +23,9 @@
 /* Header */
 
 /* Content */
+.ui.modal > .ui.card {
+  margin-top: 0;
+}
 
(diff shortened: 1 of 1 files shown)


src/styles/semantic/themes/livingstone/modules/modal.overrides
```diff
@@ -23,6 +23,9 @@
 /* Header */
 
 /* Content */
+.ui.modal > .ui.card {
+  margin-top: 0;
+}
```
davidpelayo: Just to mention the bad PR naming. I understood it as deleting loc, instead of `reset`ing the `margin-top`.

marcdavi-es: 😞

PR #188, opened 2018-06-13
TE-455 Fix server side render DOMPurify bug

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-455)

Fixes server side render DOMPurify bug

### Any other notes
Right now we get the following error at npreview-livingstone.lodgifystaging.com ...

screen shot 2018-06-13 at 16 29 49

This is because `DOMPurify.sanitize` is being called in render methods of the `Video` and `HTML` components during server side rendering.

`DOMPurify` is a browser technology and relies on browser features to work. It fails when run on the server.

By moving the call of `DOMPurify.sanitize` to `componentDidMount`, we ensure that it only gets called when mounted in the DOM, which guarantees that `DOMPurify` can access the features it needs to run.


marcdavi-es:
@davidpelayo Yeah I know what you mean. BUT I think the output is the same so it remains an effective unit test?

PR #195, opened 2018-06-18
TE-461 Dropdown fix isDisabled

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-461)

Unify options and optionsWithImages to cover a missing case where disabled wasn't checking optionsWithImages.

### Any other notes
kapture 2018-06-18 at 13 49 18

### DIFF
--- a/src/components/inputs/Dropdown/utils/getDefaultValue.js
+++ b/src/components/inputs/Dropdown/utils/getDefaultValue.js
@@ -1,8 +1,18 @@
 /**
- * If the options have images, return the first option.
- * If the options have no images, return null.
- * @param  {Object} optionsWithImages
+ * @param  {Object[]} options
+ * @param  {Boolean} hasImages
+ * @param  {Boolean} hasLabel
  * @return {String|Number|null}
  */
-export const getDefaultValue = optionsWithImages =>
-  optionsWithImages ? optionsWithImages[0].value : null;
+export const getDefaultValue = (options, hasImages, hasLabel) => {
+  /**
+   * If
+   * 1. there is at least one option, and
+   * 2. one or more of the options has an image, or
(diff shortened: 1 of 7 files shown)


src/components/inputs/Dropdown/utils/getDefaultValue.js
```diff
@@ -1,8 +1,14 @@
...
+ * @param  {Object[]} options
+ * @param  {Boolean} hasImages
+ * @param  {Boolean} hasLabel
  * @return {String|Number|null}
  */
-export const getDefaultValue = optionsWithImages =>
-  optionsWithImages ? optionsWithImages[0].value : null;
+export const getDefaultValue = (options, hasImages, hasLabel) => {
+  if (options.length) {
+    if (hasImages || !hasLabel) {
```
marcdavi-es: No nested `if` statements please.

src/components/inputs/Dropdown/utils/getDefaultValue.js
```diff
@@ -1,8 +1,14 @@
...
+ * @param  {Object[]} options
+ * @param  {Boolean} hasImages
+ * @param  {Boolean} hasLabel
  * @return {String|Number|null}
  */
-export const getDefaultValue = optionsWithImages =>
-  optionsWithImages ? optionsWithImages[0].value : null;
+export const getDefaultValue = (options, hasImages, hasLabel) => {
+  // If there are `options` and
+  // it either has images or doesn't have a label
```
marcdavi-es: This is a bit confusing as 'it' has to mean two things here. Better to format it like...

```
/**
 * If 
 * 1. there is at least one option, and
 * 2. one or more of the options has an image, or
 * 3. the Dropdown does not have a label
 */
```

PR #217, opened 2018-07-09
TE-590 Add the ability to reverse the label position

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-590)

Add the ability to render the label on the left side of the label.

### Any other notes
screen shot 2018-07-09 at 12 36 42


src/components/elements/Icon/Readme.md
```diff
@@ -141,6 +141,14 @@ Library icons
 <Icon color='black' isCircular isColorInverted />
 ```
 
+#### Reversed
+
+```jsx
+<Icon label="Search" name="search" />
+<Divider />
```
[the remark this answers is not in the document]
stephenkr: Oh true, ok I'll make the change.

src/components/elements/Icon/component.js
```diff
@@ -15,6 +15,7 @@ export const Component = ({
   isCircular,
   isColorInverted,
   isDisabled,
+  isReversed,
```
marcdavi-es: I think it's better to be more specific. `isReversed` could be interpreted as the icon itself being reversed?

Maybe `isLabelReversed` or even `isLabelLeft` ?

stephenkr: Yep good point, I think `isLabelLeft` makes more sense

src/components/elements/Icon/component.js
```diff
@@ -31,10 +32,11 @@ export const Component = ({
     // compatibility with dynamic components e.g. `ToolTip`
     {...otherProps}
   >
+    {!!label && isReversed === true && <Paragraph>{label}</Paragraph>}
```
marcdavi-es: No need for comparison here

`!!label && isReversed && ...`

src/components/elements/Icon/component.js
```diff
@@ -31,10 +32,11 @@ export const Component = ({
...
     <svg viewBox="0 0 24 24">
       <path d={getPath(name, path)} fill="currentColor" />
     </svg>
-    {!!label && <Paragraph>{label}</Paragraph>}
+    {!!label && isReversed === false && <Paragraph>{label}</Paragraph>}
```
marcdavi-es: Similarly...

No need for comparison here

`!!label && !isReversed && ...`

src/styles/semantic/themes/livingstone/elements/icon.variables
```diff
@@ -32,8 +32,9 @@
 /* Variations */
 @circularPadding: 0.2em 0;
 
-@iconLabelMarginLeft: @5px;
-@iconLabelMargin: 0 0 0 @iconLabelMarginLeft;
+@iconLabelMarginSpacing: @5px;
+@iconLabelMargin: 0 0 0 @iconLabelMarginSpacing;
+@iconReversedLabelMargin: 0 @iconLabelMarginSpacing 0 0;
```
marcdavi-es: Nice :)

src/components/elements/Icon/component.js
stephenkr: Actually i think it has to be `isLabelRight`, making the change now

src/components/elements/Icon/component.js
stephenkr: I would argue that explicit comparisons are a lot more performant than implicit, but if its the style I'm happy to change it.

https://jsperf.com/implicit-vs-explicit-js-conditionals

marcdavi-es: 🤔 Is that true (pun intended) for boolean values?

I take your point re coerced values...

stephenkr: Yep true :) I'll make the change, but I'd say V8 is optimized for this style when dealing with booleans.

Also crazy when comparing Safari and Chrome.

marcdavi-es: Hmm even for booleans explicit is (marginally) faster... https://jsperf.com/implicit-vs-explicit-js-conditionals-boolean-only

marcdavi-es: I'm happy to stick to explicit then :)

stephenkr: Oh yeah true, the JS compiler must be aware of the type so it has to do less work.

stephenkr: TBH, I think it does matter which approach is taken, the comparison is so close.

I'll stick with the style of the file.

PR #219, opened 2018-07-09
TE-594 Add the ability to add custom component definitions

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-594)

Introduces datepicker base styles as a custom definition file, all handled using the semantic style architecture.

### Any other notes
This update also removes the generic datepicker styles from the `input.overrides` and `input.variables` files and only leaves styles related to how the datepicker is styled in the input element provided by semantic-UI.

I should also note that adding overrides for this component is also possible, this means we can have theme specific styles for this component.


src/styles/semantic/definitions/elements/datepicker.less
```diff
@@ -0,0 +1,125 @@
+/*******************************
+            Theme
+*******************************/
+
+@type : 'element';
+@element : 'datepicker';
```
marcdavi-es: Do you think we ought to distinguish third party components with a new directory in `definitions` and `themes/[theme-name]`? Just to avoid mixing native semantic stuff and third party stuff?

I'm erring towards making this distinction. If you agree, these lines should read something like:

```less
@type: 'third-party-component'
@element: 'react-dates'
```

Let me know what you think.

src/styles/semantic/theme.less
```diff
@@ -29,7 +29,10 @@
 ---------------------*/
 
 /* Default */
-@import "~semantic-ui-less/@{themesFolder}/default/@{type}s/@{element}.variables";
+@import (optional) "~semantic-ui-less/@{themesFolder}/default/@{type}s/@{element}.variables";
+
+/* Custom Components */
+@import (optional) "./definitions/@{type}s/@{element}.variables";
```
marcdavi-es: I think we should duplicate the canonical Semantic UI directory structure, which would be...

(note: this assumes we go with my suggestion at https://github.com/lodgify/lodgify-ui/pull/219/files#r201052211)

```
/semantic
  /definitions
    /third-party-components
      react-dates-datepicker.less
  /themes
    /default
      /third-party-components
        react-dates-datepicker.variables
```

...as a minimum. Of course, `livingstone`-specific variables could be added at `themes/livingstone/third-party-components`.

These lines would then read 

```
/* Third party components */
@import (optional) "@{themesFolder}/default/@{type}s/@{element}.variables";
```

src/styles/semantic/themes/livingstone/elements/input.overrides
```diff
@@ -158,110 +158,6 @@
         }
       }
     }
```
marcdavi-es: I acknowledge what you say in the PR description, but I'd probably rather that everything from [line 96](https://github.com/lodgify/lodgify-ui/pull/219/files#diff-a1566383e80d9429d51788edb866a6c0L96) up to here (excluding the `textarea` stuff, obviously) is moved as well.

There will be a little duplication but I think it's better for keeping all third party stuff in one place and avoiding mixing Semantic UI native with third party stuff.

src/styles/semantic/themes/livingstone/elements/input.overrides
```diff
@@ -158,110 +158,6 @@
...
-
-          tr:not(:first-child) td.CalendarDay {
-            border-top: @dateRangePickerPickerTableCellBorder;
-          }
-        }
-      }
-    }
   }
 
   .DateRangePicker {
```
marcdavi-es: Similar to https://github.com/lodgify/lodgify-ui/pull/219/files#r201065286, I'd probably prefer the partial duplication of here to [line 397](https://github.com/lodgify/lodgify-ui/pull/219/files#diff-a1566383e80d9429d51788edb866a6c0L397) by moving this across, given the advantages of a clean boundary between semantic ui native and third party stuff.

Open to discuss this point tho.

src/styles/semantic/themes/livingstone/elements/input.variables
```diff
@@ -39,28 +39,6 @@
 @dateRangePickerDateInputPadding: (@verticalPadding - @dateRangePickerInputBorderOffset) 0;
 @dateRangePickerDateInputWidth: ~"calc(50% - @{dateRangePickerIconWidthOffset})";
```
marcdavi-es: Subject to your response to https://github.com/lodgify/lodgify-ui/pull/219/files#r201065286, these guys may also want to move across.

marcdavi-es:
Oh and before I forget - commit message should be `refactor(...` as we're not changing the api

src/styles/semantic/definitions/elements/datepicker.less
stephenkr: Yep, that makes sense. 

This would give us the flexibility to have multiple datepicker dependencies in the project if we choose.

src/styles/semantic/theme.less
stephenkr: Oh interesting, I was trying to decide where the variables file should live, i took the approach implement in the overrides folder.

But i think this makes sense, so the variables file will be specific to the chosen theme, and as long as optional is used on the import statement, the code should be safe if another theme is implemented.

src/styles/semantic/themes/livingstone/elements/input.overrides
stephenkr: Oh interesting, although this does open up and interesting topic, where should styles that effect a component used in another component go?

Does it make sense for them to exist in the `react dates datepicker` styles, or should we have another stylesheet for cross component integrations, such as `input--datepicker.less`

src/styles/semantic/themes/livingstone/elements/input.overrides
stephenkr: Yep good point, my comment above talks about this, we can discuss once in office.

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
```diff
@@ -0,0 +1,68 @@
+
+/*
+    Input specific styling
+    inherits
+        - default/input.variables
+*/
+@import (optional) "~semantic-ui-less/@{themesFolder}/default/elements/input.variables";
```
stephenkr: I had to import this file since this component is dependent on the root dependency semantic provides.

We could extract these variables out to a variable file within our project and consume it in the the input component and the datepicker component, but I'm wondering if there is any need since these types of files don't add additional CSS, they only contain variables.

src/styles/semantic/semantic.less
```diff
@@ -48,4 +48,7 @@
 & { @import "~semantic-ui-less/definitions/modules/rating.less"; }
 & { @import "~semantic-ui-less/definitions/modules/transition.less"; }
 
+/* Custom components */
```
marcdavi-es: Let's call these 

```
/* Third party components */
```

?

src/styles/semantic/theme.config
```diff
@@ -72,6 +72,12 @@
 @item       : 'livingstone';
 @statistic  : 'livingstone';
 
+/*
+  Custom Components
+  In dash case to be consistent with folder structure
+*/
```
marcdavi-es: Same as https://github.com/lodgify/lodgify-ui/pull/219/files#r201300402

I don't think we need the casing comment :)

src/styles/semantic/theme.less
```diff
@@ -29,14 +29,15 @@
...
-@import "~semantic-ui-less/@{themesFolder}/default/@{type}s/@{element}.variables";
+@import (optional) "~semantic-ui-less/@{themesFolder}/default/@{type}s/@{element}.variables";
 
 /* Packaged Theme */
 @import (optional) "@{themesFolder}/@{theme}/@{type}s/@{element}.variables";
 
 /* Site Theme */
 @import (optional) "@{siteFolder}/@{type}s/@{element}.variables";
 
+
```
marcdavi-es: Lose the new newline :P

src/styles/semantic/semantic.less
stephenkr: Yep good point, i'll change it now.

src/styles/semantic/theme.config
stephenkr: OK i'll remove it

src/styles/semantic/theme.less
stephenkr: True, fixing now.

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
marcdavi-es: I'm still not totally clear why we need this - it seems like overkill?

Can we not just identify the variables we need and declare them in  this file? I'm guessing it's

```
@verticalPadding
@inputHoverBorderColor
```

or am i missing something?

stephenkr: The issue is down to how we extend the underlying component, input. The theme.less file imports the underlying component and then utilizes those variables in the overrides file. 

Because we have styles where this component is being modified does it make sense to use the same underlying variables file or have our own version? And if so should I also expose them to the input component also?

src/styles/semantic/themes/livingstone/elements/input.variables
```diff
@@ -6,10 +6,7 @@
       Element
 --------------------*/
 
-@inputMaxWidth: 350px;
-@inputWidth: 100%;
-
-@border: @borderWidth solid transparent;
+@inputBorder: @borderWidth solid transparent;
```
marcdavi-es: I'm not sure it's safe to change the variable name here... we are relying on `@border` to modify e.g. https://github.com/Semantic-Org/Semantic-UI-LESS/blob/master/definitions/elements/input.less#L50

stephenkr: Oh good point, i'll revert. I wanted to make this variable more specific and have it live in the globals so both the input and datepicker could access it. Made sense but them couldn't progress since borderWidth only existed in the input variables file.

Fixing now.

src/styles/semantic/theme.config
```diff
@@ -72,6 +72,11 @@
 @item       : 'livingstone';
 @statistic  : 'livingstone';
 
+/*
+  Third party components
+*/
```
marcdavi-es: Single line comment for consistency please - see `/* Views */` on line 67 and others :)

stephenkr: Oh, will do, changing now.

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
```diff
@@ -0,0 +1,77 @@
+/* Imported variables */
```
marcdavi-es: This file could do with a bit of a tidy up:

1. we can de-noise a bit by removing the `dateRangePicker` prefix. In any case this is inaccurate as many of the `dateRangePicker` prefixed variables are used to style `singleDatePicker`
2. we can reorganise the file a bit better.

My suggestion attached 

[datepicker.variables.txt](https://github.com/lodgify/lodgify-ui/files/2180480/datepicker.variables.txt)

marcdavi-es: (obviously this will lead to some consequent changes in the definitions file.)

stephenkr: Good point, i'll improve the file. It was just a blind copy and paste to transfer the existing code over.

I'll clean it up.

marcdavi-es:
Nice - last thing (I promise)... 

We try to avoid uppercase in commit messages except for code entities (component or function names etc.)

Could we change commit message to something like

```
refactor(semantic): isolate datepicker as third party component
```

🙏

PR #220, opened 2018-07-09
TE-587 Create square icon SVG path

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-587)

Implements the square SVG icon path and adds examples

### Any other notes
- Fixes the icon's nested paragraph color inheritance 

image
image
image


marcdavi-es:
Nice.

Agree with @stephenkr we should avoid `!important` wherever possible. In this case, Semantic UI is twisting our arm... 💪

PR #224, opened 2018-07-11
TE-589 implement the availability widget

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-589)

Adds the availability widget to the UI library

### Any other notes
kapture 2018-07-11 at 15 11 20

screen shot 2018-07-11 at 10 11 48

The widget will jump 4 four months when the next link is clicked, on mobile the widget will jump only 1 month.

I had to also update the datepicker definition styles to support the widget, since the `CalendarMonth` is not wrapped with a picker in this widget.

The component will also assume that the selection for `All` will be included in the prop, `roomsWithImages `.

The availability of a date is handled using a function that is passed to the widget, the function will have access to the current date moment object and the selected room, this function is triggered when the calendar is rerendered and when a room is changed.

The component still requires a number of improvements that need to be applied to other components used in the widget, they include:
- `Icon`, needs to have a hover state and the ability to make the label bold, we should also have support for making the color `light`, this is not supported from the component props.
- `Dropdown`, could have an inline label next to it, instead of relying on the Grid system, we could write some CSS to better apply the positioning of the label, perhaps with flexbox or display grid.

Not important right now, but good to be aware that they will need improvements to be consistent with the provided design.

I have also requested the core dependency, React-dates, to add support for this use case in terms of displaying the months as a grid `https://github.com/airbnb/react-dates/issues/1253#issuecomment-403604203`
We cannot rely on this request being implemented, but we should keep an eye on this issue to see if the enhancement ever gets implemented.


marcdavi-es:
@stephenkr We've been using [Kap](https://getkap.co/) to record little gif screen recordings of dynamic components in action.

Saves the text description :)

src/components/general-widgets/AvailabilityWidget/Readme.md
```diff
@@ -0,0 +1,11 @@
+```jsx
+const { roomsWithImages } = require('./mock-data/options');
+
+const isDateBlocked = (moment, currentRoomValue) => moment.format('dddd') === 'Friday' && currentRoomValue === 'all';
+
+<AvailabilityWidget
```
marcdavi-es: Let's just call this `Availability`

stephenkr:
@marcdavi-es Oh very good, I'll include a gif.

src/components/general-widgets/AvailabilityWidget/Readme.md
stephenkr: Sounds good, i suppose `Widget` is unnecessary

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import { CalendarMonth, CalendarDay } from 'react-dates';
+import { BLOCKED_MODIFIER as ReactDateConstantBlocked } from 'react-dates/constants';
```
marcdavi-es: I think we ought to to keep the constant in SCREAMING_SNAKE_CASE

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+
+import 'react-dates/initialize';
+
+const ReactDateClasses = {
+  blockedDay: 'blocked-calendar',
+};
+
+/**
+ * The component for the availability widget,
+ * it will display a collection of calendars to indicate room availability
```
marcdavi-es: Note that this description is what shows up in the documentation. Follow the existing components and use something like

`The standard widget for displaying the availability of a property or room.`

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Yep ok, i'll improve the naming.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+import { GridColumn } from 'layout/GridColumn';
+import { withResponsive } from 'utils/with-responsive';
+import { Icon } from 'elements/Icon';
+import { Dropdown } from 'inputs/Dropdown';
+
+import 'react-dates/initialize';
+
+const ReactDateClasses = {
+  blockedDay: 'blocked-calendar',
+};
```
marcdavi-es: This can be simplified and should be moved to a local constants file.

```js
// constants.js

export const BLOCKED_DAY = 'blocked-calendar';

// component.js

import { BLOCKED_DAY } from './constants';

...
```

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * The component for the availability widget,
+ * it will display a collection of calendars to indicate room availability
+ * @extends {React.PureComponent}
+ * @returns {Object}
+ */
+class Component extends PureComponent {
+  state = {
+    currentRoomValue: null,
```
marcdavi-es: I don't think we need to persist the current room value? We can just expose an `onChangeRoomDropdown` function prop which passes on the value of the selected room.

It's then the consumer's responsibility to pass the right `isDateBlocked` function for the selected room.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+    super(props);
+
+    // set default value to the first item in the options array
+    if (props.roomsWithImages.length > 0) {
+      this.state = {
+        currentRoomValue: props.roomsWithImages[0].value,
+        startDate: this.state.startDate,
+      };
+    }
+  }
```
marcdavi-es: We don't need the constructor method.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+
+  getMonths = (date, isUserOnMobile) => {
+    return Array(isUserOnMobile ? 1 : 4)
+      .fill(date)
+      .map((item, index) => item.clone().add(index, 'month'));
+  };
```
marcdavi-es: This guy is a pure function so he can be isolated as a local util.

Also - I think a more specific name would help... maybe `getMonthsToDisplay`

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+
+  onNextMonthChange = isUserOnMobile =>
```
marcdavi-es: `handleClickNextMonth` is more specific and consistent with the project.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+    this.setState({
+      startDate: this.state.startDate
+        .clone()
+        .add(isUserOnMobile ? 1 : 4, 'month'),
+    });
+
+  onPreviousMonthChange = isUserOnMobile =>
```
marcdavi-es: `handleClickPreviousMonth`

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+    this.setState({
+      startDate: this.state.startDate
+        .clone()
+        .subtract(isUserOnMobile ? 1 : 4, 'month'),
```
marcdavi-es: The logic modifying the moment can be isolated as a pure function local util with a unit test against it.

```js
startDate: getPreviousMonth(this.state.startDate)
```

Also true for the next month.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+    });
+
+  renderCalendarDay = props => {
+    const modifiers = new Set([]);
+    const { isDateBlockedWithRoom } = this.props;
+    const { currentRoomValue } = this.state;
+
+    // Trigger method to determine if the current date is blocked.
```
marcdavi-es: I don't think we need this comment.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+
+    return <CalendarDay {...props} modifiers={modifiers} />;
+  };
+
+  reloadCalendarOnRoomSelection = (evt, currentRoomValue) => {
+    this.setState({
+      currentRoomValue,
+    });
+    this.forceUpdate();
+  };
```
marcdavi-es: I think we can lose this guy if we're not persisting `currentRoomValue`

stephenkr:
Another design requirement is to add the weekday mapping above the month calendars.

I can do this in another PR because it will involve bumping the dependency to the next version and updating some styles to remain compatible.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Yep good idea.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Ok so the consumer will keep the awareness of the current room instead of the availability component?

stephenkr: Sounds good since less state will exist in the component

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+                  name="arrow left"
+                  onClick={() => this.onPreviousMonthChange(isUserOnMobile)}
+                />
+              </GridColumn>
+              <GridColumn textAlign="right" width={6}>
+                <Icon
+                  isLabelLeft
+                  label="Next"
+                  name="arrow right"
+                  onClick={() => this.onNextMonthChange(isUserOnMobile)}
```
marcdavi-es: For this (and the one for previous) we can avoid the inline function declaration by just passing the class method and getting `isUserOnMobile` from props in the method...

```js
...

onClickNextMonth = () => this.setState({ 
  startDate: getPreviousMonth(this.state.startDate, this.props.isUserOnMobile) 
});

...

onClick={this.onClickNextMonth}
```

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Yep good point.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Good idea, i'll extract it out.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+  onDayClick: () => {},
+};
+
+Component.propTypes = {
+  /**
+   * A function that is passed to determine if the current date should be blocked, the function is passed a moment object
+   * @param {Moment} momentObj
+   * @param {Object} currentRoomObject
+   */
+  isDateBlockedWithRoom: PropTypes.func,
```
marcdavi-es: Let's make the API consistent with e.g. `DateRangePicker`...


```
  /**
   * A function called for each day to be displayed. Returning true blocks that day in the calendar.
   * @param   {Moment}  day - The day to test.
   * @returns {Boolean}     - Is the day blocked.
   */
  getIsDayBlocked: PropTypes.func,
```

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+   * Is the user on a mobile device.
+   * Provided by `withResponsive` so ignored in the styleguide.
+   * @ignore
+   */
+  isUserOnMobile: PropTypes.bool.isRequired,
+  /**
+   * When a day is clicked in one of the calendars, this method will trigger
+   * @param {Moment} momentObj
+   */
+  onDayClick: PropTypes.func,
```
marcdavi-es: We don't need to expose this - there's no functionality on clicking a day.

src/components/general-widgets/AvailabilityWidget/component.js
```diff
@@ -0,0 +1,237 @@
...
+      /** The visible text for the option. */
+      text: PropTypes.string.isRequired,
+      /** The underlying value for the option. */
+      value: PropTypes.oneOfType([
+        PropTypes.bool,
+        PropTypes.number,
+        PropTypes.string,
+      ]),
+    })
+  ).isRequired,
```
marcdavi-es: Let's make the API consistent with e.g. `Contact`...

```js
/** The options which the user can select for the room field. */
  roomOptions: PropTypes.arrayOf(
    ...
  ),
```

Also - I'm not sure images are required so we should hard code this into the prop name

src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
```diff
@@ -64,58 +52,69 @@
...
+          color: @pickerDayBlockedFontColor;
         }
       }
     }
+
+    tr:not(:first-child) td.CalendarDay {
+      border-top: @pickerTableCellBorder;
+    }
   }
 }
```
marcdavi-es: Nice.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Yep i agree, so it contains the type of event that will be triggered.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Fixing now.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: So i need the calendars to rerender on room change, this will ensure that happens.

Perhaps there is a more efficient way to do so.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Yep looks cleaner.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Will do.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Yep true, i was more thinking about a possible future use case. But since we dont need it, i'll remove it.

src/components/general-widgets/AvailabilityWidget/component.js
stephenkr: Images are a requirement based on the type of dropdown we are using.

I'll update the prop to be more consistent with how the `Contact`.. components are implemented.

stephenkr:
I have a few more improvements to apply to this PR, please don't approve and merge yet.

src/components/property-page-widgets/Availability/Readme.md
```diff
@@ -0,0 +1,13 @@
+```jsx
+const { roomsWithImages } = require('./mock-data/options');
+
+let currentRoomSelection = roomsWithImages[0].value;
+
+const isDateBlocked = (moment) => moment.format('dddd') === 'Friday' && currentRoomSelection === 'all';
```
marcdavi-es: Let's change this guy to `isDayBlocked` for consistency with the prop name?

src/components/property-page-widgets/Availability/component.js
```diff
@@ -0,0 +1,219 @@
...
+import { getNextStartDate } from './utils/getNextStartDate';
+import { getPreviousStartDate } from './utils/getPreviousStartDate';
+import { BLOCKED_DAY_CLASS } from './constants';
+
+import 'react-dates/initialize';
+
+/**
+ * The standard widget for displaying the availability of a room.
+ * @extends {React.PureComponent}
+ * @returns {Object}
```
marcdavi-es: No `@returns` tag for classes :)

src/components/property-page-widgets/Availability/component.js
```diff
@@ -0,0 +1,219 @@
...
+   */
+  isUserOnMobile: PropTypes.bool.isRequired,
+  /**
+   * A function that triggered when a room is changed in the room dropdown
+   * @param {Event} event
+   * @param {String} currentRoomValue
+   */
+  onChangeRoomDropdown: PropTypes.func,
+  /* The room dropdown options, on selection of a room, the availability calendar will update */
+  roomsWithImages: PropTypes.PropTypes.arrayOf(
```
marcdavi-es: I think this should be `roomOptions` or `roomOptionsWithImages`, for consistency with other components which have dropdowns.

src/components/property-page-widgets/Availability/Readme.md
stephenkr: yeah ok, i'll change it.

src/components/property-page-widgets/Availability/component.js
```diff
@@ -0,0 +1,219 @@
...
+   * @ignore
+   */
+  isUserOnMobile: PropTypes.bool.isRequired,
+  /**
+   * A function that triggered when a room is changed in the room dropdown
+   * @param {Event} event
+   * @param {String} currentRoomValue
+   */
+  onChangeRoomDropdown: PropTypes.func,
+  /* The room dropdown options, on selection of a room, the availability calendar will update */
```
marcdavi-es: `/** The options which the user can select for the room field. */`

src/components/property-page-widgets/Availability/component.js
stephenkr: Fixing now.

src/components/property-page-widgets/Availability/utils/getMonthsToDisplay.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @param  {Moment} date
+ * @param  {Boolean} isUserOnMobile
+ * @return {Moment[]}
+ */
+export const getMonthsToDisplay = (date, isUserOnMobile) => {
+  return Array(isUserOnMobile ? 1 : 4)
+    .fill(date)
+    .map((item, index) => item.clone().add(index, 'month'));
+};
```
marcdavi-es: Nice.

src/components/property-page-widgets/Availability/component.js
stephenkr: I'll go with `roomOptionsWithImages`, fixing now.

src/components/property-page-widgets/Availability/component.js
```diff
@@ -0,0 +1,218 @@
...
+              </Grid>
+            </GridColumn>
+            <GridColumn
+              only="tablet computer"
+              textAlign="right"
+              verticalAlignContent="middle"
+              width={5}
+            >
+              <Icon
+                color="grey"
```
Jonnygwi: this colour should be light grey. `color="light grey"`

stephenkr: Yep good point, i had `light grey` in the color attribute but started getting an error in the console. I think the component needs to be updated to support the light variation in the prop types.

I mentioned this above in the description of the PR `we should also have support for making the color light, this is not supported from the component props.`

src/components/property-page-widgets/Availability/component.spec.js
```diff
@@ -0,0 +1,416 @@
+// add spec to ensure `blocked-calendar` will always exist in the react-dates dependency
```
Jonnygwi: Is this meant to be here?

stephenkr: Oh true, removing now.

src/components/property-page-widgets/Availability/utils/getNextStartDate.spec.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe('getMonthsToDisplay', () => {
+  it('should return the correct start date if not on mobile', () => {
+    const actual = getNextStartDate(today, false);
+    expect(
+      today
+        .clone()
+        .add(4, 'month')
+        .isSame(actual, 'date')
+    ).toBeTruthy();
```
Jonnygwi: We should avoid using `.toBeTruthy` and replace it with `.toBe(true)`

https://vincenttunru.com/toBeTruthy-vs-toBe-true/

src/components/property-page-widgets/Availability/utils/getNextStartDate.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('should return the correct start date if on mobile', () => {
+    const actual = getNextStartDate(today, true);
+    expect(
+      today
+        .clone()
+        .add(1, 'month')
+        .isSame(actual, 'date')
+    ).toBeTruthy();
```
Jonnygwi: same as https://github.com/lodgify/lodgify-ui/pull/224/files#r201704165

src/components/property-page-widgets/Availability/utils/getPreviousStartDate.spec.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe('getMonthsToDisplay', () => {
+  it('should return the correct start date if not on mobile', () => {
+    const actual = getPreviousStartDate(today, false);
+    expect(
+      today
+        .clone()
+        .subtract(4, 'month')
+        .isSame(actual, 'date')
+    ).toBeTruthy();
```
Jonnygwi: same as https://github.com/lodgify/lodgify-ui/pull/224/files#r201704165

src/components/property-page-widgets/Availability/utils/getPreviousStartDate.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('should return the correct start date if on mobile', () => {
+    const actual = getPreviousStartDate(today, true);
+    expect(
+      today
+        .clone()
+        .subtract(1, 'month')
+        .isSame(actual, 'date')
+    ).toBeTruthy();
```
Jonnygwi: same as https://github.com/lodgify/lodgify-ui/pull/224/files#r201704165

src/components/property-page-widgets/Availability/utils/getMonthsToDisplay.spec.js
```diff
@@ -0,0 +1,26 @@
...
+describe('getMonthsToDisplay', () => {
+  it('should return the correct number of months is not on mobile', () => {
+    const actual = getMonthsToDisplay(today, false);
+    expect(actual).toHaveLength(4);
+  });
+
+  it('should return the months in sequential order', () => {
+    const actual = getMonthsToDisplay(today, false);
+
+    expect(actual[0].isSame(today, 'date')).toBeTruthy();
```
Jonnygwi: same as https://github.com/lodgify/lodgify-ui/pull/224/files#r201704165

src/components/property-page-widgets/Availability/utils/getMonthsToDisplay.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    expect(
+      actual[3].isSame(today.clone().add(3, 'months'), 'date')
+    ).toBeTruthy();
```
Jonnygwi: same as https://github.com/lodgify/lodgify-ui/pull/224/files#r201704165

src/components/property-page-widgets/Availability/utils/getNextStartDate.spec.js
stephenkr: Yep good point, I do wish there was a matcher called `toBeOnlyTrue`.

I'll clean up the code now.

src/components/property-page-widgets/Availability/Readme.md
```diff
@@ -0,0 +1,13 @@
+```jsx
+const { roomOptionsWithImages } = require('./mock-data/options');
+
+let currentRoomSelection = roomOptionsWithImages[0].value;
```
Jonnygwi: We shouldn't be using let statements as it means that we are relying on side effects. In this case we could turn this into a pure function but as this is just an example I would delete this and the function on line 11 and replace it for a `console.log`. This is only an example so doing this functionality isn't necessary

stephenkr: So the example is to demonstrate that the availability component will render the blocked dates depending on the current room selected and the current date being passed.

We decided to move this logic out of the component and have the parent take ownership over room selection awareness.

I do agree that we should avoid mutable variables, but this is only an example use case.

PR #228, opened 2018-07-12
TE-601 Update ReactDates to v17

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-601)

Upgrades the dependency, `react-dates`, from v16 to v17

### Any other notes
A number of breaking changes have been introduced in this upgrade.
1. Default styling has been removed from custom navigation buttons, because of this we need to add our own CSS to resolve this issue
2. The datepicker height is now computed using the number of weeks rather than the computed height, so the datepicker assumes the height of the cells will always be the same and due to our custom stying, a white space will appear after the dates. So the fix is to expand the height of the table cells to respect this assumption.

__Preview of the datepickers with the new version__

screen shot 2018-07-12 at 11 06 26
screen shot 2018-07-12 at 11 06 13

### DIFF
--- a/src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
+++ b/src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
@@ -33,6 +33,23 @@
       top: @pickerNavigationButtonTopSpace;
     }
 
+    // add styling for buttons
+    .DayPickerNavigation button:not(.DayPickerNavigation_button__horizontalDefault) {
+      position: absolute;
+      top: @pickerButtonTop;
+      line-height: @pickerButtonLineHeight;
+      border-radius: @pickerButtonBorderRadius;
+      padding: @pickerButtonPadding;
+
+      &:first-child {
+        left: @pickerButtonFirstLeft;
+      }
+
+      &:last-child {
+        right: @pickerButtonLastRight;
+      }
+    }
+
     .DayPicker_weekHeader {
       color: inherit;
       padding: @pickerTablePadding;
--- a/src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
+++ b/src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
@@ -43,13 +43,21 @@
 
 @pickerTablePadding: 0 @15px;
 @pickerTableCellBorder: @4px solid @pureWhite;
-@pickerTableCellHeight: 38px !important;
+@pickerTableCellHeight: 48px !important;
 @pickerTableCellWidth: 48px !important;
 
 @pickerDayHoveredBackground: @lightPrimaryColor;
 @pickerDaySelectedBackground: @primaryColor;
 @pickerDayBlockedBackground: @greyTwo;
 @pickerDayBlockedFontColor: @greyFive;
+
+/* Buttons */
+@pickerButtonTop: 1.7rem;
+@pickerButtonLineHeight: 0.78em;
+@pickerButtonBorderRadius: @3px;
+@pickerButtonPadding: @6px @9px;
+@pickerButtonFirstLeft: @20px;
+@pickerButtonLastRight: @20px;
 
 /*-------------------
         States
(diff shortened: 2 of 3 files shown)


src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
```diff
@@ -33,6 +33,23 @@
       top: @pickerNavigationButtonTopSpace;
     }
 
+    // add styling for buttons
+    .DayPickerNavigation button:not(.DayPickerNavigation_button__horizontalDefault) {
+      position: absolute;
+      top: 1.7rem;
```
Jonnygwi: Let's turn all of these values to variables

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
```diff
@@ -43,7 +43,7 @@
 
 @pickerTablePadding: 0 @15px;
 @pickerTableCellBorder: @4px solid @pureWhite;
-@pickerTableCellHeight: 38px !important;
+@pickerTableCellHeight: 48px !important;
```
Jonnygwi: Same as https://github.com/lodgify/lodgify-ui/pull/228/files#r201972253

src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
stephenkr: Sure, just the dimensions?

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
stephenkr: It's already a variable though? Or do you mean have a shared var for the width and height of the cell since those values are the same?

Jonnygwi: Ahh yeah sorry my mistake

stephenkr: No problem :D

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
```diff
@@ -43,14 +43,22 @@
...
 @pickerDayBlockedBackground: @greyTwo;
 @pickerDayBlockedFontColor: @greyFive;
 
+/* Buttons */
+@pickerButtonTop: 1.7rem;
+@pickerButtonLineHeight: 0.78em;
+@pickerButtonBorderRadius: 3px;
+@pickerButtonPadding: 6px 9px;
+@pickerButtonFirstLeft: 20px;
+@pickerButtonLastRight: 20px;
```
Jonnygwi: We should use the global variables provided by semantic-ui `@XXpx`. Check the `node_modules/semantic-ui-less/themes/default/globals/site.variables`

stephenkr: Oh ok, I'll make the change.

Why though just out of curiosity? doesn't it seem a bit insane to have @20px === 20px

src/styles/semantic/themes/default/third-party-components/react-dates-datepicker.variables
```diff
@@ -43,14 +43,27 @@
...
 @pickerDaySelectedBackground: @primaryColor;
 @pickerDayBlockedBackground: @greyTwo;
 @pickerDayBlockedFontColor: @greyFive;
 
+/* Buttons */
+
+// These values are used to better position the navigation buttons in the react datepicker
+// Custom navigation elements no longer inherit the built in styles, we now must provide them.
+// The original values can be found here, some values have been modified to better fit
+// https://github.com/airbnb/react-dates/blob/83ec316167b499ce939058973da0aff30e70d178/src/components/DayPickerNavigation.jsx#L239
```
marcdavi-es: I don't think we need all these comments - it's just doing what code does :)

stephenkr: Ok, I'll remove them.

marcdavi-es: lol

PR #230, opened 2018-07-12
TE-602 Improve overall test coverage

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-602)

Adds the missing unit tests for `Header`, `getPluralString` and `Footer`.

### DIFF
--- a/src/components/collections/Footer/utils/getGroupedNavigationItems.spec.js
+++ b/src/components/collections/Footer/utils/getGroupedNavigationItems.spec.js
@@ -123,6 +123,10 @@ describe('getGroupedNavigationItems', () => {
           },
         ],
       },
+      {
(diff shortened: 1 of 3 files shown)


src/components/collections/Footer/utils/getGroupedNavigationItems.spec.js
```diff
@@ -123,6 +123,10 @@ describe('getGroupedNavigationItems', () => {
           },
         ],
       },
+      {
```
stephenkr: Just thinking, but would it be worth explaining that if the object is empty, it will be dropped and the last item will be nested under a subItems array?

Looks a bit misleading when looking at the before and after, it looks like the last item is merged with the second last.

PR #232, opened 2018-07-13
Launch script to use IDE debugger in VS Code

stephenkr:
No ticket

This will add a launch config to the project so we can use the built-in VS code debugger when running Jest tests

### Any other notes
Using the VS code debugger will give us access to using breakpoints in our test code and it will allow us to easily run specs from the IDE.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,30 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+    "type": "node",
+    "request": "launch",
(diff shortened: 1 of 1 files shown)


marcdavi-es:
I don't agree with adding discretionary config to the source code. 

If a developer needs editor specific config, it can live in their local project directory.

stephenkr:
Yep ok, that makes sense.

We can update the .gitignore file to ignore the .vscode folder.

PR #233, opened 2018-07-13
TE-604 Add week day mapping to Availability calendars

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-604)

Adds weekday mapping to Availability calendars

### Any other notes
I wish the dependency React Dates handled this for us, but it looks like we need to construct the weekday mapping header ourselves.

I have used the same classes they have applied in the datepicker components they provide, this will help us inherit any styles they provide to better allow the header template to sit naturally in the availability component.

I had to also introduce a new dependency to add support for the moment object in PropTypes.

screen shot 2018-07-13 at 15 42 55

### DIFF
--- a/src/components/property-page-widgets/Availability/utils/renderMonthHeader.js
+++ b/src/components/property-page-widgets/Availability/utils/renderMonthHeader.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import moment from 'moment';
+import momentPropTypes from 'react-moment-proptypes';
+
+import { WEEKDAY_LENGTH } from '../constants';
+
+const buildWeekDayMappingMarkup = () => (
+  <ul className="DayPicker_weekHeader_ul">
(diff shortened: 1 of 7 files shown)


src/components/property-page-widgets/Availability/utils/renderMonthHeader.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import moment from 'moment';
+import momentPropTypes from 'react-moment-proptypes';
+
+const buildWeekDayMappingMarkup = () => (
+  <ul className="DayPicker_weekHeader_ul">
+    {Array(7)
```
Jonnygwi: We should probably pull this `7` out into a constant

stephenkr: Ok will do

PR #234, opened 2018-07-13
TE-610 Extend Amenities

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-610)

Reworks the Amenities component to allow for amenities to always be categorised 

### Any other notes
- Expose labelWeight prop in `Icon` 

#### Mobile view
image

#### Modal
image

### DIFF
--- a/src/components/elements/Icon/component.js
+++ b/src/components/elements/Icon/component.js
@@ -17,6 +17,7 @@ export const Component = ({
   isDisabled,
   isLabelLeft,
   label,
+  labelWeight,
   name,
   path,
   size,
@@ -32,11 +33,13 @@ export const Component = ({
     // compatibility with dynamic components e.g. `ToolTip`
     {...otherProps}
   >
-    {!!label && isLabelLeft && <Paragraph>{label}</Paragraph>}
+    {!!label &&
+      isLabelLeft && <Paragraph weight={labelWeight}>{label}</Paragraph>}
     <svg viewBox="0 0 24 24">
       <path d={getPath(name, path)} fill="currentColor" />
     </svg>
-    {!!label && !isLabelLeft && <Paragraph>{label}</Paragraph>}
+    {!!label &&
+      !isLabelLeft && <Paragraph weight={labelWeight}>{label}</Paragraph>}
   </i>
 );
 
@@ -49,6 +52,7 @@ Component.defaultProps = {
   isDisabled: false,
   isLabelLeft: false,
   label: null,
+  labelWeight: null,
   name: null,
   path: null,
   size: null,
@@ -82,6 +86,8 @@ Component.propTypes = {
   isLabelLeft: PropTypes.bool,
   /** A visible label to display with the icon. */
   label: PropTypes.string,
+  /** The weight of the label. */
+  labelWeight: PropTypes.oneOf(['heavy', 'light']),
   /** The name of the icon to display. Takes priority over `props.path`. */
   name: PropTypes.oneOf([
     'arrow down',
--- a/src/components/elements/Icon/component.spec.js
+++ b/src/components/elements/Icon/component.spec.js
@@ -85,9 +85,21 @@ describe('<Icon />', () => {
   });
 
   describe('the `Paragraph` component', () => {
+    const getIconWithLabel = props =>
+      getIcon({ label: 'someLabel', ...props }).find(Paragraph);
+
     it('should have the right children', () => {
-      const wrapper = getIcon({ label: 'someLabel' }).find(Paragraph);
+      const wrapper = getIconWithLabel();
       expectComponentToHaveChildren(wrapper, 'someLabel');
+    });
+
+    describe('if `props.labelWeight` is passed', () => {
+      it('should have the right props', () => {
+        const wrapper = getIconWithLabel({
+          labelWeight: 'heavy',
+        });
+        expectComponentToHaveProps(wrapper, { weight: 'heavy' });
+      });
     });
   });
 
--- a/src/components/property-page-widgets/Amenities/Readme.md
+++ b/src/components/property-page-widgets/Amenities/Readme.md
@@ -1,15 +1,40 @@
 ```jsx
-const { sixAmenities } = require('./mock-data/amenities');
+const amenities = [
+  {
+    name: 'Cooking',
+    iconName: 'leaf',
+    items: ['Toaster', 'Microwave', 'Coffee Machine'],
+  },
+  {
+    name: 'Bathroom & Laundry',
+    iconName: 'paw',
+    items: ['Bidet', 'Hair Dryer', 'Iron & Board'],
+  },
+  {
+    name: 'Spa & Gym',
+    iconName: 'leaf',
+    items: ['Ping-Pong Table', 'Pool Table', 'Foosball'],
+  },
+  {
+    name: 'Outside',
+    iconName: 'sun',
+    items: ['Tennis Court'],
+  },
+  {
+    name: 'Miscellaneous',
+    iconName: 'coffee',
+    items: [
+      'Rice Steamer',
+      'Hot Tub',
+      'Fold-Away Bed',
+      'Fireplace',
+      'Steam Sauna',
+      'Fitness-Room',
+      'Parking',
+      'Laundry Service',
+    ],
+  },
+];
 
-<Amenities amenities={sixAmenities} />
-```
-
-### Content
-
-#### Extra icons
-
-```jsx
-const { twelveAmenities } = require('./mock-data/amenities');
-
-<Amenities amenities={twelveAmenities} />
+<Amenities amenities={amenities} />
 ```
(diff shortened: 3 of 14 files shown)


src/components/elements/Icon/component.js
```diff
@@ -82,6 +86,7 @@ Component.propTypes = {
   isLabelLeft: PropTypes.bool,
   /** A visible label to display with the icon. */
   label: PropTypes.string,
+  labelWeight: PropTypes.string,
```
marcdavi-es: Please add JSDoc comment for this prop.

src/components/property-page-widgets/Amenities/component.js
```diff
@@ -1,49 +1,36 @@
...
  * The standard widget for displaying the amenities of a property.
  * @returns {Object}
  */
-const Component = ({ amenities, isUserOnMobile }) => (
-  <Grid>
+const Component = ({ amenityCategories, isUserOnMobile }) => (
+  <Grid stackable>
     <GridColumn width={12}>
-      <Heading>Amenities</Heading>
+      <Heading>Property Amenities</Heading>
```
marcdavi-es: Why are we including 'Property' ?

src/components/property-page-widgets/Amenities/component.js
```diff
@@ -53,18 +40,15 @@ const Component = ({ amenities, isUserOnMobile }) => (
 Component.displayName = 'Amenities';
 
 Component.propTypes = {
-  /** The amenities to display as icons. */
-  amenities: PropTypes.arrayOf(
+  /** The amenity categories. */
+  amenityCategories: PropTypes.arrayOf(
```
marcdavi-es: I think we can leave this as `amenities`

src/components/property-page-widgets/Amenities/utils/getCategoryMarkup.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+import { buildKeyFromStrings } from 'utils/build-key-from-strings';
+import { GridColumn } from 'layout/GridColumn';
+import { GridRow } from 'layout/GridRow';
+import { Icon } from 'elements/Icon';
+
+import { getFormattedAmenities } from './getFormattedAmenities';
+
+export const getCategoryMarkup = (category, keyString) => (
```
marcdavi-es: JSDoc tags.

src/components/property-page-widgets/Amenities/utils/getFormattedAmenities.js
```diff
@@ -0,0 +1,13 @@
+export const getFormattedAmenities = amenities =>
```
marcdavi-es: JSDoc tags.

src/components/elements/Icon/component.js
```diff
@@ -17,6 +17,7 @@ export const Component = ({
   isDisabled,
   isLabelLeft,
   label,
```
stephenkr: I think if you have another prop called `labelWeight`, perhaps you should call the `label` prop something more specific, such as `labelText`

src/components/property-page-widgets/Amenities/utils/getFormattedAmenities.js
```diff
@@ -0,0 +1,13 @@
...
+    // Is penultimate item in array
+    if (index === amenities.length - 2) {
+      return [...accumulator, `${amenity} and `];
+    }
+    // Is last item in array
+    if (index === amenities.length - 1) {
+      return [...accumulator, amenity];
+    }
+
+    return [...accumulator, `${amenity}, `];
```
stephenkr: Just wondering, but would a map be suitable for this?

src/components/elements/Icon/component.js
Jonnygwi: @stephenkr this sounds good. This change would need to be across the whole project as there are other components using the `label` prop name so for consistency they also need to change. Could you create a tech task ticket for this?

stephenkr: Yep i'll create a ticket :)

stephenkr: Ticket created. https://youtrack.lodgify.net/agiles/86-18/87-449?issue=TE-614

src/components/elements/Icon/component.js
```diff
@@ -32,11 +33,13 @@ export const Component = ({
     // compatibility with dynamic components e.g. `ToolTip`
     {...otherProps}
   >
-    {!!label && isLabelLeft && <Paragraph>{label}</Paragraph>}
+    {!!label &&
```
stephenkr: Minor, but i always think its cleaner to use `Boolean` instead of the double explanation marks.

src/components/elements/Icon/component.spec.js
```diff
@@ -85,10 +85,22 @@ describe('<Icon />', () => {
...
+      getIcon({ label: 'someLabel', ...props }).find(Paragraph);
+
     it('should have the right children', () => {
-      const wrapper = getIcon({ label: 'someLabel' }).find(Paragraph);
+      const wrapper = getIconWithLabel();
       expectComponentToHaveChildren(wrapper, 'someLabel');
     });
+
+    describe('if `props.labelWeight` is passed', () => {
+      it('should have the right props if` ', () => {
```
stephenkr: Looks like something is missing in the spec description.

src/components/property-page-widgets/Amenities/Readme.md
```diff
@@ -1,15 +1,40 @@
...
-
-### Content
-
-#### Extra icons
-
-```jsx
-const { twelveAmenities } = require('./mock-data/amenities');
-
-<Amenities amenities={twelveAmenities} />
+<Amenities amenities={amenities} />
```
stephenkr: Minor, but looks funny to see `amenities` so many times :D 

I wonder if it would make more sense to call the prop `items`? or `sections` depending on how the object array is being used.

marcdavi-es: In general we reuse the name of the component for the primary prop. See e.g. `KeyFacts` etc.

src/components/property-page-widgets/Amenities/component.spec.js
```diff
@@ -1,35 +1,33 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import { Responsive } from 'semantic-ui-react';
+import { Responsive } from 'semantic-ui-react/dist/commonjs';
```
stephenkr: Why not just `semantic-ui-react`?

stephenkr: Oh you could group this with the other semantic-ui-react import below.

src/components/property-page-widgets/Amenities/component.js
```diff
@@ -1,49 +1,36 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Modal as SemanticModal } from 'semantic-ui-react/dist/commonjs';
```
stephenkr: Also the import here.

stephenkr: Comment below.

src/components/property-page-widgets/Amenities/component.js
```diff
@@ -53,18 +40,15 @@ const Component = ({ amenities, isUserOnMobile }) => (
...
+      /** The name of the icon to display for the category. */
       iconName: PropTypes.string.isRequired,
-      /** Is the amenity disabled. */
-      isDisabled: PropTypes.bool,
-      /** A visible label to display for the amenity. */
-      label: PropTypes.string.isRequired,
+      /** The list of amenities */
+      items: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
+      /** The category name */
+      text: PropTypes.string.isRequired,
```
stephenkr: Would it make more sense to call this `name` rather than text? Since the use will be as a name?

src/components/property-page-widgets/Amenities/component.spec.js
marcdavi-es: Oh yeah this is kind of weird 🤔 - what's this in aid of?

src/components/property-page-widgets/Amenities/Readme.md
stephenkr: Oh ok, good to know :)

src/components/elements/Icon/component.js
Jonnygwi: I think both `!!label` and `Boolean(label)` are clear so it's not a big deal for me but if it's a big deal for you maybe you can add it to the conventions docs and people can vote on it.

stephenkr: Oh ok, i think using `Boolean` is more explicit, although i see in the Airbnb JS standards the double explanation is the best approach.

https://github.com/airbnb/javascript#coercion--booleans

stephenkr: Let's stick with the double explanation, ignore my comment above.

PR #235, opened 2018-07-16
TE-613 Get availability component spec coverage to 100%

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-613)

Get availability component spec coverage to 100%

### DIFF
--- a/src/components/property-page-widgets/Availability/component.spec.js
+++ b/src/components/property-page-widgets/Availability/component.spec.js
@@ -253,6 +253,12 @@ describe('<Availability />', () => {
         wrapper.simulate('change', {});
         expect(onChangeRoomDropdown).toBeCalled();
       });
+
+      it('should have `Function.prototype` as the default value for `props.onChangeRoomDropdown`', () => {
+        const wrapper = getWrappedAvailability();
+        const props = wrapper.instance().props;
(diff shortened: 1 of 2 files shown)


src/components/property-page-widgets/Availability/component.spec.js
```diff
@@ -253,6 +253,12 @@ describe('<Availability />', () => {
         wrapper.simulate('change', {});
         expect(onChangeRoomDropdown).toBeCalled();
       });
+
+      it('ensure the default prop of `onChangeRoomDropdown` returns nothing', () => {
+        const wrapper = getWrappedAvailability();
+        const props = wrapper.instance().props;
+        expect(props.onChangeRoomDropdown()).toBe(undefined);
+      });
```
marcdavi-es: Ah - I missed this in the earlier PR.

For no-op functions as `defaultProps`, we've been using `Function.prototype`. Can we change that in `component.js` please?

Then the test here is 

```
it('should have `Function.prototype` as the default value for `props.onChangeRoomDropdown`', ...)
```

stephenkr: Yes of course, i'll apply the change.

PR #236, opened 2018-07-16
TE-611 Amenities responsive

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-611)

Adds the responsive layout for the extended Amenities component

### Any other notes
- Changed example icons

kapture 2018-07-16 at 15 01 58

### DIFF
--- a/src/components/property-page-widgets/Amenities/utils/getCategoryMarkup.js
+++ b/src/components/property-page-widgets/Amenities/utils/getCategoryMarkup.js
@@ -1,27 +1,29 @@
 import React from 'react';
 
+import { Paragraph } from 'typography/Paragraph';
 import { buildKeyFromStrings } from 'utils/build-key-from-strings';
 import { GridColumn } from 'layout/GridColumn';
-import { GridRow } from 'layout/GridRow';
 import { Icon } from 'elements/Icon';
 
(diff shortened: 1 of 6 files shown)


src/components/property-page-widgets/Amenities/utils/getCategoryMarkup.js
```diff
@@ -1,27 +1,29 @@
...
-        labelWeight="heavy"
-        name={category.iconName}
-      />
-    </GridColumn>
-    <GridColumn>{getFormattedAmenityItems(category.items)}</GridColumn>
-  </GridRow>
+export const getCategoryMarkup = (category, index, isUserOnMobile = false) => (
+  <GridColumn
+    key={buildKeyFromStrings(category.name, index)}
+    width={isUserOnMobile ? 4 : 12}
```
stephenkr: You can use the prop `computer="12"` and `mobile="4"` instead, may remove the need for the `isUserOnMobile ` prop.

Jonnygwi: This `GridColumn` doesn't always want to have `width={4}` when the user is on mobile. The modal uses the same markup helper and it needs a `width={12}` when in mobile.

PR #237, opened 2018-07-17
TE-619 Remove react-docgen dependency conflict

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-619)

Removes react-docgen from `package.json` and locks react-styleguidist to a version below 7.1.0

### Any other notes
This fixes the issue of components that aren't using `class ... extends React.Component` from not being rendered


package.json
```diff
@@ -84,7 +84,6 @@
     "npm": "^5.8.0",
     "npm-run-all": "^4.1.2",
     "prettier": "^1.12.1",
-    "react-docgen": "^2.20.0",
```
marcdavi-es: OK I think this is fine - I believe it used to be a `peerDepdency` of `react-styleguidist` but is now included in their `dependencies`.

BUT - this isn't a complete solution. Let's fix the problem completely in one PR.

Jonnygwi: Yep, we are working on that now.

PR #240, opened 2018-07-17
TE-615 Remove `.isRequired` from `bedroomsNumber` on `FeaturedProperty`

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-615)

Remove `.isRequired` from `bedroomsNumber` on `FeaturedProperty`

Note that the ticket says remove the prop altogether. I don't think this is necessary - we can leave it on here as we/others may or may not consume the prop later.

### DIFF
--- a/src/components/general-widgets/FeaturedProperty/component.js
+++ b/src/components/general-widgets/FeaturedProperty/component.js
@@ -51,12 +51,13 @@ export const Component = ({
 Component.displayName = 'FeaturedProperty';
 
 Component.defaultProps = {
+  bedroomsNumber: null,
   imageAlternativeText: '',
 };
 
 Component.propTypes = {
   /** The number of available bedrooms at the property. */
-  bedroomsNumber: PropTypes.number.isRequired,
+  bedroomsNumber: PropTypes.number,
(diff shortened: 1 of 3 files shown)


src/components/general-widgets/FeaturedProperty/component.js
```diff
@@ -51,12 +51,13 @@ export const Component = ({
 Component.displayName = 'FeaturedProperty';
 
 Component.defaultProps = {
+  bedroomsNumber: null,
```
stephenkr: Is it right to define this as null?

Would 0 be more suitable since the variable name expects it to be a number?

marcdavi-es: Hmmm. 

- I think `0` is confusing as it's a valid number (a featured property could validly have zero rooms).
- In general we've been using `null` when we want to explicitly pass a non-value as the default prop value.

stephenkr: Yep that's a good point, I see your point on 0 being a valid value for the bedrooms number.

PR #243, opened 2018-07-17
TE-621 Fixes the modal in the RoomType component

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-621)

Fixes the modal in the RoomType component

### Any other notes
Make items optional in Amenities so it works in the RoomType component and updates the RoomType props to be compatible with the Amenities component.

One of the bugs was introduced when the items prop was made a requirement in the Amenities component.
The existing component, RoomType, will pass in an array of items without the required sub-array, even though this is a requirement in the Amenities component. The fix was to make this sub array optional in the Amenities component.

The other bug includes the prop `labelText` in the RoomType component, it will pass this in the amenities array into the Amenities component, which the RoomType consumes. This is not valid since the Amenities component expects the `labelText` value to be defined as `name`.

This change will fix those two bugs.

#### Steps to reproduce bug
1. navigate to https://lodgify.github.io/lodgify-ui/#/Property%20page%20widgets/RoomType
2. Click on the "More Info" link
3. Observe the result

It is expected to see the modal rendered with additional information

__Broken version__
kapture 2018-07-17 at 17 18 59

__Working version after fix__
kapture 2018-07-17 at 17 20 25

### DIFF
--- a/src/components/property-page-widgets/Amenities/component.js
+++ b/src/components/property-page-widgets/Amenities/component.js
@@ -48,7 +48,7 @@ Component.propTypes = {
       /** The name of the icon to display for the category. */
       iconName: PropTypes.string.isRequired,
       /** The list of amenities */
-      items: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
+      items: PropTypes.arrayOf(PropTypes.string.isRequired),
       /** The category name */
       name: PropTypes.string.isRequired,
     })
--- a/src/components/property-page-widgets/Amenities/utils/getCategoryMarkup.js
+++ b/src/components/property-page-widgets/Amenities/utils/getCategoryMarkup.js
@@ -24,6 +24,8 @@ export const getCategoryMarkup = (category, index, isFullWidth = false) => (
       labelWeight="heavy"
       name={category.iconName}
     />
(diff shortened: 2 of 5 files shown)


marcdavi-es:
@stephenkr It's not clear to me from the PR description or the [ticket](https://youtrack.lodgify.net/issue/TE-621) what the 'before' and 'after' states are for this change. 

Can you show what is broken and what the intended fix looks like, please?

stephenkr:
@marcdavi-es The description of this bug has been updated.

Jonnygwi:
With the new changes to Amenities I'm not sure if this component should have this behaviour. @jams676 can you confirm this is the intended behaviour of this component? For example the modal says 'Property Amenities' but this component is more specifically about a room etc

Jonnygwi:
@stephenkr I've spoken to @jams676 and he says that the component should not have this behaviour. The Amenities need to have the structure in the screenshot. Only one extension needs to be done and that's on the Amenities title. In this case it should be `Room Amenities`.

image

stephenkr:
Thanks @Jonnygwi 

I'll make the necessary changes now.

PR #244, opened 2018-07-18
TE-617 Extend Contact component

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-617)

Extends the Contact component to have the ability to only show the necessary dropdowns

### Any other notes
- Fixed a issue in the Form component when `React.createElement` and `React.cloneElement` was trying to clone or create `null` or `undefined`. 

kapture 2018-07-18 at 11 43 06

### DIFF
--- a/src/components/collections/Form/component.js
+++ b/src/components/collections/Form/component.js
@@ -23,12 +23,14 @@ export class Component extends PureComponent {
   };
 
   cloneFormGroup = formGroup =>
+    formGroup &&
     React.cloneElement(formGroup, {
       children: Children.map(formGroup.props.children, this.cloneInput),
       widths: 'equal',
     });
 
   cloneInput = input =>
+    input &&
     React.createElement(Form.Field, {
       children: React.cloneElement(input, {
         onChange: (name, value) => {
@@ -53,7 +55,7 @@ export class Component extends PureComponent {
             {Children.map(
               children,
               child =>
-                child.type === Form.Group
+                child && child.type === Form.Group
                   ? this.cloneFormGroup(child)
                   : this.cloneInput(child)
             )}
--- a/src/components/general-widgets/Contact/Readme.md
+++ b/src/components/general-widgets/Contact/Readme.md
@@ -2,10 +2,62 @@
 const mockCaptcha = require('./mock-data/signupcode.jpeg');
 const {
   propertyOptions,
+  roomOptions,
 } = require('./mock-data/options');
 
 <Contact
   captchaInputImage={mockCaptcha}
   propertyOptions={propertyOptions}
+  roomOptions={roomOptions}
+/>
+```
+
+### Content
+
+
+#### Dropdowns
+
+No dropdown will be rendered if there is no options prop.
+
+```jsx
+const mockCaptcha = require('./mock-data/signupcode.jpeg');
+const {
+  roomOptions,
+} = require('./mock-data/options');
+
+<Contact
+  captchaInputImage={mockCaptcha}
+  roomOptions={roomOptions}
+/>
+```
+
+A disabled dropdown will be rendered if the options array is empty.
+
+```jsx
+const mockCaptcha = require('./mock-data/signupcode.jpeg');
+const {
+  roomOptions,
+} = require('./mock-data/options');
+
+<Contact
+  captchaInputImage={mockCaptcha}
+  propertyOptions={[]}
+  roomOptions={roomOptions}
+/>
+```
+
+A dropdown will be rendered if the options array is populated.
+
+```jsx
+const mockCaptcha = require('./mock-data/signupcode.jpeg');
+const {
+  propertyOptions,
+  roomOptions,
+} = require('./mock-data/options');
+
+<Contact
+  captchaInputImage={mockCaptcha}
+  propertyOptions={propertyOptions}
+  roomOptions={roomOptions}
 />
 ```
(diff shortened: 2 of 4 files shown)


src/components/collections/Form/component.js
```diff
@@ -53,7 +55,7 @@ export class Component extends PureComponent {
             {Children.map(
               children,
               child =>
-                child.type === Form.Group
+                child && child.type === Form.Group
```
marcdavi-es: Are you sure this is necessary?

src/components/general-widgets/Contact/Readme.md
```diff
@@ -2,10 +2,60 @@
...
+
+```jsx
+const mockCaptcha = require('./mock-data/signupcode.jpeg');
+const {
+  roomOptions,
+} = require('./mock-data/options');
+
+<Contact
+  captchaInputImage={mockCaptcha}
+  roomOptions={[]}
```
marcdavi-es: Let's keep the `roomOptions` prop consistent in these examples so that that difference is easy to identify. i.e. `roomOptions` is always present and populated, `propertyOptions` changes.

So, here we want 

```jsx
<Contact
  captchaInputImage={mockCaptcha}
  propertyOptions={[]}
  roomOptions={roomOptions}
```

src/components/general-widgets/Contact/Readme.md
```diff
@@ -2,10 +2,60 @@
...
+/>
+```
+
+A dropdown will be rendered if the options are in the array.
```
marcdavi-es: `...if the options array is populated.`

src/components/general-widgets/Contact/component.js
```diff
@@ -59,6 +65,7 @@ Component.displayName = 'Contact';
 Component.defaultProps = {
   onChangeProperty: Function.prototype,
   onSubmit: Function.prototype,
+  propertyOptions: undefined,
```
stephenkr: We should probably decide on a standard value for default variables like these, either `undefined` or `null`

src/components/general-widgets/Contact/Readme.md
```diff
@@ -2,10 +2,60 @@
...
+```jsx
+const mockCaptcha = require('./mock-data/signupcode.jpeg');
+const {
+  propertyOptions,
+} = require('./mock-data/options');
+
+<Contact
+  captchaInputImage={mockCaptcha}
+  propertyOptions={propertyOptions}
+  roomOptions={[]}
```
marcdavi-es: following from https://github.com/lodgify/lodgify-ui/pull/244/files#r203323293 ...

```jsx
<Contact
  captchaInputImage={mockCaptcha}
  propertyOptions={propertyOptions}
  roomOptions={roomOptions}
```

src/components/general-widgets/Contact/component.js
```diff
@@ -37,15 +37,21 @@ export const Component = ({
...
-    <InputGroup>
-      <Dropdown
-        label="Property"
-        name="property"
-        onChange={onChangeProperty}
-        options={propertyOptions}
-      />
-      <Dropdown label="Room" name="room" options={roomOptions} />
-    </InputGroup>
+    {(!!roomOptions || !!propertyOptions) && (
```
marcdavi-es: This should be `!!roomOptions &&  !!propertyOptions`

src/components/general-widgets/Contact/component.js
marcdavi-es: use `null`

src/components/general-widgets/Contact/component.js
```diff
@@ -59,6 +65,7 @@ Component.displayName = 'Contact';
 Component.defaultProps = {
   onChangeProperty: Function.prototype,
   onSubmit: Function.prototype,
+  propertyOptions: undefined,
   roomOptions: undefined,
```
marcdavi-es: use `null`

src/components/collections/Form/component.js
stephenkr: Also doesn't a child element in a loop need a key?

src/components/general-widgets/Contact/component.spec.js
```diff
@@ -144,39 +143,89 @@ describe('<Contact />', () => {
...
-    it('should have the right props', () => {
-      const wrapper = getForm()
-        .find(InputGroup)
-        .at(2)
-        .childAt(1);
-      expectComponentToHaveProps(wrapper, {
-        label: 'Room',
-        name: 'room',
-        options: options.roomOptions,
+  describe('if `propertyOptions` and `roomOptions` are passed', () => {
```
marcdavi-es: Need to test the reverse case i.e. `if neither .... are passed`

src/components/collections/Form/component.js
Jonnygwi: This is needed because the `type` is not a property of `null`

src/components/general-widgets/Contact/component.js
stephenkr: Might also be best to change the prop `roomOptions` to `null` too to standardize the props.

src/components/general-widgets/Contact/component.spec.js
Jonnygwi: There is a case for the starting on line 44

src/components/collections/Form/component.js
```diff
@@ -53,7 +55,7 @@ export class Component extends PureComponent {
             {Children.map(
               children,
               child =>
-                child.type === Form.Group
```
stephenkr: > Also doesn't a child element in a loop need a key?

Just a comment above.

PR #245, opened 2018-07-18
TE-624 Set heading text from prop in Amenities

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-624)

Add prop to set the heading text in the component

### Any other notes

screen shot 2018-07-18 at 16 01 34
Preview of component with no heading (top example), and a component with the heading defined(bottom example)


src/components/property-page-widgets/Amenities/component.js
```diff
@@ -53,6 +59,7 @@ Component.propTypes = {
       name: PropTypes.string.isRequired,
     })
   ).isRequired,
+  headingText: PropTypes.string,
```
marcdavi-es: Needs a comment so it gets documented in styleguidist. Please copy and paste (with necessary tweaks) from some other component which has `headingText`

stephenkr: Yep true, adding now.

Jonnygwi: Ahh good spot! 👀

PR #248, opened 2018-07-19
TE-640 Add fullscreen option to modal

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-640)

Add fullscreen option to modal

kapture 2018-07-19 at 18 15 56

### DIFF
--- a/src/styles/semantic/themes/livingstone/modules/modal.overrides
+++ b/src/styles/semantic/themes/livingstone/modules/modal.overrides
@@ -36,6 +36,17 @@
 /* Mobile Positions */
 
 /* Responsive Widths */
+.ui.fullscreen.modal {
+  bottom: 0;
+  height: @fullScreenHeight;
+  position: absolute;
+  top: 0;
+
+  .content {
+    height: 100%;
+    overflow: auto;
+  }
+}
 
 /* Coupling */
 
--- a/src/styles/semantic/themes/livingstone/modules/modal.variables
+++ b/src/styles/semantic/themes/livingstone/modules/modal.variables
@@ -25,6 +25,9 @@
 /* Mobile Positions */
 
 /* Responsive Widths */
+@fullScreenWidth: 100%;
+// Fixes the bug I describe at https://stackoverflow.com/questions/41453780/semantic-react-modal-weird-height-behavior/51426377#51426377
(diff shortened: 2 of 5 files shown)


src/styles/semantic/themes/livingstone/modules/modal.overrides
```diff
@@ -36,6 +36,13 @@
 /* Mobile Positions */
 
 /* Responsive Widths */
+.ui.fullscreen.modal {
+  bottom: 0;
+  margin: 0;
+  margin-bottom: @fullScreenMarginBottom;
+  position: absolute;
+  top: 0;
+}
```
stephenkr: I think we need to account for content expanding outside of the modal container,

Add the following css to the `.content` of the modal
```
    overflow: auto;
    max-height: 100vh;
```

src/styles/semantic/themes/livingstone/modules/modal.overrides
```diff
@@ -36,6 +36,13 @@
 /* Mobile Positions */
 
 /* Responsive Widths */
+.ui.fullscreen.modal {
+  bottom: 0;
+  margin: 0;
+  margin-bottom: @fullScreenMarginBottom;
```
stephenkr: You can just use the shorthand `margin: 0 0 @ fullScreenMarginBottom;`

src/styles/semantic/themes/livingstone/modules/modal.variables
```diff
@@ -25,6 +25,9 @@
 /* Mobile Positions */
 
 /* Responsive Widths */
+@fullScreenWidth: 100%;
+// Fixes the bug I describe at https://stackoverflow.com/questions/41453780/semantic-react-modal-weird-height-behavior/51426377#51426377
```
stephenkr: The fix looks good, but an alternative approach could be to set a max height of `max-height: calc(100% - 1px);` or `max-height: calc(100vh - 1px);`

marcdavi-es: Yep on reflection I think `height: calc(100% - 1px);` is the least hacky of the available hacks :)

stephenkr: Yep that looks good 👍

PR #250, opened 2018-07-23
TE-674 Add pinterest icon

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-674)

Adds Pinterest icon

### Any other notes
Change the name of the `google` Icon to `google plus`

image


src/components/elements/Icon/Readme.md
```diff
@@ -118,10 +118,12 @@ Library icons
 <Heading size="small">Social media</Heading>
 <Icon labelText="facebook" name="facebook" />
 <Divider />
-<Icon labelText="google" name="google" />
+<Icon labelText="google plus" name="google plus" />
```
stephenkr: I think it should be `googleplus`

src/components/elements/Icon/component.js
```diff
@@ -114,7 +114,7 @@ Component.propTypes = {
     'double bed',
     'facebook',
     'fire',
-    'google',
+    'google plus',
```
stephenkr: Same as above. I think it should be `googleplus`

src/components/elements/Icon/Readme.md
marcdavi-es: In future - let's not lose time on these kinds of discussions. Fall back on what Semantic UI does... https://react.semantic-ui.com/elements/icon

In this case... `google plus`

stephenkr: Ok, sounds good.

PR #251, opened 2018-07-23
TE-630 `Header` responsive

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-630)

Adds responsive structure and styles for `Header`

kapture 2018-07-23 at 17 36 53

### DIFF
--- a/src/components/collections/Header/component.js
+++ b/src/components/collections/Header/component.js
@@ -1,12 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Menu, Image } from 'semantic-ui-react';
-import { size } from 'lodash';
+import { Menu } from 'semantic-ui-react';
 
-import { buildKeyFromStrings } from 'utils/build-key-from-strings';
-import { Heading } from 'typography/Heading';
-import { Submenu } from 'elements/Submenu';
-import { Button } from 'elements/Button';
+import { withResponsive } from 'utils/with-responsive';
+
+import { getLogoMarkup } from './utils/getLogoMarkup';
+import { getMobileMenuMarkup } from './utils/getMobileMenuMarkup';
+import { getDesktopMenuMarkup } from './utils/getDesktopMenuMarkup';
 
 /**
  * A header displays a logo, grouped navigation items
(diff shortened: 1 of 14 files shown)


src/components/collections/Header/utils/getMobileMenuMarkup.js
```diff
@@ -0,0 +1,79 @@
...
+        locationOptions={searchBarLocationOptions}
+      />
+    </Menu.Item>
+    <Menu.Item>
+      <Modal isFullscreen trigger={<Icon name="bars" />}>
+        <Menu text vertical>
+          {getLogoMarkup(logoSrc, logoText)}
+          {navigationItems.map(
+            ({ subItems, text, href }, index) =>
+              size(subItems) ? (
```
stephenkr: Can you use `.length` instead? since `subItems` is expected to be an array.

stephenkr: Also, i think its cleaner to have `subItems.length > 0`

src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -189,6 +209,40 @@
...
 
   /* Vertical */
+  .ui.modal &.vertical {
+    width: @verticalMenuInModalWidth;
+
+    > .item:first-child {
+      padding-top: 0;
+    }
+
+    > .item:not(:first-child) {
```
stephenkr: You can group these two selectors, such as:
```
> .item {
 &:first-child {
  ...
 }
 &:not(:first-child) {
  ...
 }
}
```

src/styles/semantic/themes/livingstone/collections/menu.variables
```diff
@@ -106,6 +108,16 @@
 ---------------*/
 
 /* Vertical */
+@verticalMenuInModalWidthOffset: 2 * @textVerticalMenuHorizontalMargin;
+@verticalMenuInModalWidth: ~"calc(100% - @{verticalMenuInModalWidthOffset})";
+
+@verticalMenuInModalItemLineHeight: @30px;
+@verticalMenuInModalItemPaddingLeft: @10px;
+@verticalMenuInModalItemBorderBottom: @1px solid @greyFour;
```
stephenkr: I'm curious on the mindset being these dimension variables,

Would `@1px` ever not be `1px` and if it was, should it be called `@1px`, so why do these variables exist?

src/components/collections/Header/utils/getMobileMenuMarkup.js
marcdavi-es: `size` covers the case where `subItems` is `null` or `undefined`

src/components/collections/Header/component.js
```diff
@@ -95,10 +67,41 @@ Component.propTypes = {
...
   primaryCTA: PropTypes.shape({
     /** The href url for the call to action. */
     href: PropTypes.string.isRequired,
     /** The  visible text for the call to action. */
     text: PropTypes.string.isRequired,
   }),
+  /** The options which the user can select in the guests field in the search bar. */
+  // eslint-disable-next-line react/no-unused-prop-types
+  searchBarGuestsOptions: PropTypes.arrayOf(
+    PropTypes.shape({
```
stephenkr: I wonder if its worth extracting these Prop shapes into something more reusable? Ive seen this pattern a lot.

src/components/collections/Header/utils/getMobileMenuMarkup.js
stephenkr: You could alternatively do `(subItems || []).length`, one less dependency :)

marcdavi-es: Yep there are lots of alternative ways. 

In this case (following other components in lodgify-ui) we chose to use a ready-to-go util rather than reimplementing it and taking the unit test burden on ourselves.

src/styles/semantic/themes/livingstone/collections/menu.variables
marcdavi-es: We do this to tie values to the `@emSize`. See...

https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/globals/site.variables#L537

I'm not certain we'll actually need these values to be modified with `emSize` but we chose to go this way and we're sticking to it for consistency.

I don't think there's any downside.

stephenkr: Ok, I think we need to be careful using the `em` dimension for element sizing, ideally `em` should be used for relative values around typography.

Ideally, we should rely on using `rem` since it only respects the root font size.

But for consistency, its fine. Just looks strange when first looking at it.

src/components/collections/Header/component.js
stephenkr: Its `a suggested refactor across the project`

I'll create an issue for the board.

src/components/collections/Header/utils/getMobileMenuMarkup.js
stephenkr: Ok fair enough

src/styles/semantic/themes/livingstone/collections/menu.variables
marcdavi-es: @stephenkr  This does evaluate to `rem` - if you check the link you'll see it.

stephenkr: Yep, it looks good.

PR #252, opened 2018-07-24
TE-677 ESlint empty line

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-677)

Adds the ESlint config to enforce empty lines after variable definitions and runs `npm run lint:js:fix`

[Rule](https://eslint.org/docs/rules/newline-after-var#require-or-disallow-an-empty-line-after-variable-declarations-newline-after-var)


marcdavi-es:
Oh wow - I thought we were already using https://www.npmjs.com/package/eslint-config-lodgify

Workflow should be:
1. implement `eslint-config-lodgify` in this project
1. suggest rules as PRs into that repo.

PR #259, opened 2018-07-31
TE-721 Icon name mapping

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-721)

Add and expose icon name mapping


src/components/elements/Icon/constants.js
```diff
@@ -0,0 +1,103 @@
...
+  SQUARE: 'square',
+  SUN: 'sun',
+  TRAIN: 'train',
+  TWITTER: 'twitter',
+  USERS: 'users',
+  WHEELCHAIR: 'wheelchair',
+  YOUTUBE: 'youtube',
+};
+
+export const ICON_NAMES_LIST = [
```
marcdavi-es: This is asking for duplication and inconsistency problems.

Do this programatically...

```js
export const ICON_NAMES_LIST = Object.values(ICON_NAMES);
```

src/components/inputs/Dropdown/component.spec.js
```diff
@@ -40,7 +40,8 @@ describe('<Dropdown />', () => {
       const wrapper = getSemanticDropdown();
       expectComponentToHaveProps(wrapper, {
         defaultValue: null,
-        icon: <Icon name="caret down" />,
+        icon: <Icon name={ICON_NAMES.CARET_DOWN} />,
+        icon: <Icon name={ICON_NAMES.CARET_DOWN} />,
```
stephenkr: I think you can remove the extra item here.

PR #260, opened 2018-08-01
TE-716 Add extra icon paths

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-716)

Adds paths for, port, underground, wifi, credit card, no children, parking, cooking, entertainment, heating and laundry

kapture 2018-08-01 at 14 43 42


src/components/elements/Icon/Readme.md
```diff
@@ -84,24 +84,43 @@ Library icons
 <Divider />
 <Icon labelText="coffee" name="coffee" />
 <Divider />
+<Icon labelText="cooking" name="cooking" />
```
stephenkr: I think you can make this into a loop and get the icons from the constants file.

stephenkr: Although its probably better to write each icon since this will be documented and can be a useful reference for a reader.

PR #270, opened 2018-08-10
TE-798: Add width as a prop for GridColumn

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-798)

Adds width as a prop for GridColumn

### DIFF
--- a/src/components/layout/GridColumn/component.js
+++ b/src/components/layout/GridColumn/component.js
@@ -16,9 +16,12 @@ Component.displayName = 'GridColumn';
 
 Component.defaultProps = {
   verticalAlignContent: 'top',
+  width: 12,
 };
 
 Component.propTypes = {
   /** Vertically align the content of the column to the bottom, middle or top. */
   verticalAlignContent: PropTypes.oneOf(['bottom', 'middle', 'top']),
+  /** The width of the column */
+  width: PropTypes.number,
 };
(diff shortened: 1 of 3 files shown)


src/components/layout/GridColumn/component.js
```diff
@@ -16,9 +16,12 @@ Component.displayName = 'GridColumn';
...
 Component.defaultProps = {
   verticalAlignContent: 'top',
+  width: 12,
 };
 
 Component.propTypes = {
   /** Vertically align the content of the column to the bottom, middle or top. */
   verticalAlignContent: PropTypes.oneOf(['bottom', 'middle', 'top']),
+  /** The width of the column */
+  width: PropTypes.number,
```
marcdavi-es: Use `PropTypes.oneOf` 

We only want to allow numbers `1` to `12`

src/components/layout/GridColumn/component.spec.js
```diff
@@ -24,6 +24,7 @@ describe('<GridColumn />', () => {
 
       expectComponentToHaveProps(wrapper, {
         verticalAlign: expect.any(String),
+        width: expect.any(Number),
```
marcdavi-es: You can be specific, right? It should be `12` since that's the default prop value.

src/components/layout/GridColumn/component.js
stephenkr: Will do.

src/components/layout/GridColumn/component.spec.js
stephenkr: Yep true.

PR #273, opened 2018-08-17
TE-839 Expose strings props on `Contact`

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-839)

Exposes strings props on `Contact`.

### Any other notes

Please review this carefully as the agreed solution will be followed across all components.

I considered two approaches for exposing strings props...

1. Expose a single `strings` prop, being an object containing each of the necessary strings; and
1. Expose a single prop for each of the necessary strings.

I opted for 2., for the following reasons...

- Flat string props are great for rerendering on update. There are no complications associated with object identity.
- It makes use of the top level React `defaultProps` API with no intermediate steps. This is great both for simplicity of code and for props documentation in Styleguidist. 
- It is the simplest solution, requiring the minimum change.

One contentious point is the isolation of default strings as a module at `src/utils/default-strings`. This is good for the code - we avoid duplication and spelling mistakes - but it looks a bit wonky in the documentation... 

screen shot 2018-08-17 at 12 23 34

I'm comfortable with the trade off. Considering our primary audience for the documentation is developers, I think the UPPERCASED identifiers are easily understood as constants.

I am merging this without review in order to unblock myself on [TE-833](https://youtrack.lodgify.net/issue/TE-833) but I am open to comments and amendments.

### DIFF
--- a/src/components/collections/Form/utils/getFormChild.js
+++ b/src/components/collections/Form/utils/getFormChild.js
@@ -18,7 +18,7 @@ export const getFormChild = (child, parent) => {
   if (child.type === Form.Group) {
     return React.cloneElement(child, {
       children: Children.map(child.props.children, nestedChild =>
-        getClonedInput(nestedChild, parent)
+        getFormChild(nestedChild, parent)
       ),
       widths: 'equal',
     });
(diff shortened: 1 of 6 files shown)


Jonnygwi:
Nice

PR #282, opened 2018-08-21
TE-890: Expose string props for Promotion

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-890)

Expose the string props for `Promotion`

__Preview__
screen shot 2018-08-21 at 15 56 12

__:hover__
screen shot 2018-08-21 at 15 56 08


src/components/general-widgets/Promotion/component.js
```diff
@@ -96,8 +109,14 @@ Component.propTypes = {
   backgroundImage: PropTypes.string.isRequired,
   /** The discount amount to be displayed */
   discountAmount: PropTypes.string.isRequired,
+  /** The above text for the discount amount  */
```
Jonnygwi: What does the above text mean?

stephenkr: Since you have asked i'll need to rethink this comment.

The text that is defined here will appear above the discount amount, normally it is `Save up too` and then the amount is shown.

Since it is not a header, i didn't want to call it `headingText`, since it will exist in a paragraph tag, so I tried to be as descriptive as I could.

I suppose i can rename the comment to be: `The label text to show above the discount amount`, what do you think @Jonnygwi?

Jonnygwi: Maybe a slight reorder makes it sound more natural -> `/** The text to display above the discount amount. */`

stephenkr: Yep looks clean, i'll update the comment

PR #283, opened 2018-08-21
TE-891: Expose string props for PropertyPictures

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-891)

Expose the string props for `PropertyPictures`

__Preview__
screen shot 2018-08-21 at 16 23 38


stephenkr:
@Jonnygwi Just thinking, but do you think the `propertyPictures` prefix on the string props is a bit unnecessary?

What if we just have `headingText` and `linkText` for the string props? I think it looks cleaner.

Jonnygwi:
It's not a bad idea. I think `headingText` is fine as `PropertyPictures` is already in context

PR #284, opened 2018-08-21
TE-893: Expose string prop for Rules

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-893)

Exposes the string prop for `Rules`

__Preview__
screen shot 2018-08-21 at 17 08 54


src/components/property-page-widgets/Rules/component.js
```diff
@@ -43,11 +49,17 @@ export const Component = ({ checkInTime, checkOutTime, rules }) => (
...
+Component.defaultProps = {
+  headingText: HOUSE_RULES,
+};
+
 Component.propTypes = {
   /** The propery check-in time. */
   checkInTime: PropTypes.string.isRequired,
   /** The propery check-out time. */
   checkOutTime: PropTypes.string.isRequired,
+  /** The heading text */
```
Jonnygwi: `/** The heading text */` <-- this is a little lazy. Something a little more descriptive would be better. `/** The text for the Rules heading */`

stephenkr: Sure i'll make the change.

PR #290, opened 2018-08-27
TE-921: Remove default strings from utility in OwnerLogin

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-921)

Removes default strings from forgot password utility in `OwnerLogin`


stephenkr:
doesnt appear to be needed, resolved in another PR

PR #298, opened 2018-08-28
TE-927:  expose string props for Rates

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-927)

Expose string props for `Rates`

__Preview__
screen shot 2018-08-28 at 14 12 28


src/components/property-page-widgets/Rates/component.js
```diff
@@ -115,6 +133,8 @@ Component.propTypes = {
   ).isRequired,
   /** The headings for each column of the table */
   rateHeadings: PropTypes.arrayOf(PropTypes.string).isRequired,
+  /** The heading text for the room type input */
+  roomTypeHeadingText: PropTypes.string,
```
marcdavi-es: This is more of a label than a heading, no?

I suggest `roomTypeInputLabel`

src/components/property-page-widgets/Rates/component.js
```diff
@@ -98,6 +114,8 @@ Component.propTypes = {
   onChangeCurrency: PropTypes.func,
   /** A function called when the room type value changes. */
   onChangeRoomType: PropTypes.func,
+  /** The text to display for the price per extra guest */
+  pricePerExtraText: PropTypes.string,
```
marcdavi-es: This creates inconsistency.

We have `rateCategories.costPerExtraGuest` already.

This should be `costPerExtraGuestLabel`

src/utils/default-strings/constants.js
```diff
@@ -48,4 +49,5 @@ export const TIME = 'Time';
 export const TIME_ZONE = 'Time Zone';
 export const USE_COUPON_CODE = 'Use the coupon code';
 export const VIEW_MORE = 'View more';
+export const VIEW_RATE_INFORMATION_FOR = 'View Rate Information for:';
```
Jonnygwi: This should not have the colon. That needs to be extracted out into a helper. `getStringWithColonSuffix` already exists.

src/components/property-page-widgets/Rates/component.js
stephenkr: I'll fix this.

src/components/property-page-widgets/Rates/utils/buildPricePerExtraGuestString.js
```diff
@@ -1,6 +1,7 @@
 /**
  * @param {String} cost
+ * @param {String} pricePerExtraText
  * @return {String}
  */
-export const buildPricePerExtraGuestString = cost =>
-  `${cost} / Price Per extra per.`;
+export const buildPricePerExtraGuestString = (cost, pricePerExtraText) =>
+  `${cost} / ${pricePerExtraText}`;
```
Jonnygwi: Do you mean to remove the full stop?

src/utils/default-strings/constants.js
stephenkr: Yep true, i'll rely on a util to append the colon

src/components/property-page-widgets/Rates/utils/buildPricePerExtraGuestString.js
stephenkr: I'll add it back.

PR #306, opened 2018-08-31
TE-955: Add styles for featured cards

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-955)

adds styling for the featured properties such as:
1. Sets the height of the feature property cards to 155px
2. Updates the spacing between the star icons is 4px

Please refer to the ticket for more information.

__Preview of featured property__
screen shot 2018-08-31 at 12 01 31

__Preview of feature room type__
screen shot 2018-08-31 at 12 01 55

---

__Before, Preview of featured property__
screen shot 2018-08-31 at 12 06 39

### DIFF
--- a/src/styles/semantic/themes/livingstone/views/card.overrides
+++ b/src/styles/semantic/themes/livingstone/views/card.overrides
@@ -83,6 +83,14 @@
     }
   }
 
+  /* Featured */
+  &.has-featured {
+
+    img.ui.image {
+      height: @featuredImageHeight;
+    }
+  }
+
   /*-------------------
       Variations
 --------------------*/
--- a/src/components/general-widgets/FeaturedProperty/component.js
+++ b/src/components/general-widgets/FeaturedProperty/component.js
@@ -23,7 +23,7 @@ export const Component = ({
   propertyUrl,
   ratingNumber,
 }) => (
-  <Card href={propertyUrl}>
+  <Card className="has-featured" href={propertyUrl}>
     <Image alt={imageAlternativeText} src={imageUrl} />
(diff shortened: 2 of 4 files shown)


src/styles/semantic/themes/livingstone/views/card.overrides
```diff
@@ -83,6 +83,18 @@
     }
   }
 
+  /* Featured */
+  &.has-featured {
+
+    img {
```
marcdavi-es: Let's use a more specific selector to avoid unintended conflicts.

```css
img.ui.image {
```

src/styles/semantic/themes/livingstone/views/card.overrides
```diff
@@ -83,6 +83,18 @@
...
+      height: @featuredImageHeight;
+    }
+
+    .ui.rating .icon {
+      width: @featuredIconWidth;
+    }
```
marcdavi-es: Check in with Julio but I believe this is relevant to all ratings? If so, we can add this to the ratings styles.

src/styles/semantic/themes/livingstone/views/card.overrides
stephenkr: Sure i'll make the selector more specific.

src/styles/semantic/themes/livingstone/views/card.overrides
stephenkr: The ticket is specific to the featured properties.

We can create another ticket to handle all reviews and then make the selector for the ratings above more inclusive.

PR #309, opened 2018-09-03
TE-958: Set spacing between icons to 4px for Rating

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-958)

Sets the rating icon to 14px for any component that consumes the `Rating` component, this will set the spacing between each icon to 4px

__Before__
screen shot 2018-09-03 at 10 56 35

__After__
screen shot 2018-09-03 at 10 56 55

### DIFF
--- a/src/styles/semantic/themes/livingstone/modules/rating.variables
+++ b/src/styles/semantic/themes/livingstone/modules/rating.variables
@@ -6,7 +6,7 @@
 
 @marginLeft: @5px;
 
-@iconWidth: @20px;
+@iconWidth: @14px;
 
 @siblingSpanMarginRight: 0.25em;
 @siblingSpanFloat: left;


src/styles/semantic/themes/livingstone/modules/rating.variables
```diff
@@ -6,7 +6,7 @@
 
 @marginLeft: @5px;
 
-@iconWidth: @20px;
+@iconWidth: @14px;
```
Jonnygwi: Is this going to effect any of the components using the same variables?

marcdavi-es: I don't think anything else consumes `rating.variables`

stephenkr: This variable will only affect the rating component.

There is another variable called `@iconWidth` in the `input.variables` file, but these two variables do not affect each other.

Jonnygwi: Perfection

PR #310, opened 2018-09-03
TE-960: remove border and shadow from Menu if in Header

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-X)

Removes the box shadow and border from the menu if it is used in `Header` and has the class `borderless`

- The `text` prop will add the class `text` to the menu, more info here https://react.semantic-ui.com/collections/menu/#types-text

__Before__
screen shot 2018-09-03 at 11 06 49

__After__
screen shot 2018-09-03 at 11 07 00

### DIFF
--- a/src/components/collections/Header/component.js
+++ b/src/components/collections/Header/component.js
@@ -15,7 +15,7 @@ import { getDesktopMenuMarkup } from './utils/getDesktopMenuMarkup';
  * @return {Object}
  */
 const Component = props => (
-  <Menu borderless className="is-header">
+  <Menu borderless className="is-header" text>
     {getLogoMarkup(props.logoSrc, props.logoText)}
     <Menu.Menu position="right">
       {props.isUserOnMobile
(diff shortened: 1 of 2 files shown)


src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -168,6 +168,11 @@
   /* stylelint-disable-next-line no-duplicate-selectors */
   &.is-header {
 
+    &.borderless {
+      border: none;
+      box-shadow: none;
+    }
```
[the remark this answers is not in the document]
stephenkr: I can use an existing variable to solve this problem, from what I can see the root of the issue is that borderless class is not doing what it is intended to do.

I'll clean this up.

src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -168,6 +168,10 @@
   /* stylelint-disable-next-line no-duplicate-selectors */
   &.is-header {
 
+    &.text {
+      margin: 0;
+    }
```
marcdavi-es: I think you can use the `@textMenuMargin` rather than writing more styles.

stephenkr: I didn't modify this variable since this may affect other implementations of the text menu, I wanted to restrict this change to the header.

marcdavi-es: I don't think there are any other implementations of `Menu text`

PR #311, opened 2018-09-03
TE-946: import semantic Container and add overrides

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-946)

Implements the Semantic Container into the Lodgify UI container and applies the container dimensions

### Any other notes
- Handles the container relationship with the grid when the container is applied to the grid row
- Updates tablet breakpoint to 670px to be consistent with breakpoints across components
- Also the grid columns now have an inner padding of `2rem`

__Box sizing__
screen shot 2018-09-03 at 09 20 40

__Preview of the container with text__
screen shot 2018-09-03 at 09 31 33
> Please note that 1120px is `1040px + 40px + 40px`

__Preview of tablet container with text__
screen shot 2018-09-03 at 09 31 54
> 940px is this case is 100% of the parent container

### The container when screen width is less than 670px

screen shot 2018-09-03 at 12 21 23

screen shot 2018-09-03 at 14 29 39

### The container with the grid

The intended behaviour of the container with the grid is for the left and right side padding of the grid row to have a padding of 40px, this includes removing the side padding of the first and last columns in the grid row.

screen shot 2018-09-03 at 10 07 17

screen shot 2018-09-03 at 12 20 17
screen shot 2018-09-03 at 12 19 40

screen shot 2018-09-03 at 10 07 23


src/components/layout/Container/component.js
```diff
@@ -0,0 +1,24 @@
...
+
+Component.displayName = 'Container';
+
+Component.defaultProps = {
+  textAlign: 'left',
+};
+
+Component.propTypes = {
+  /** The text align for nested children. */
+  textAlign: PropTypes.oneOf(['left', 'center', 'right', 'justified']),
```
marcdavi-es: What's the current use case for `textAlign`?

src/styles/semantic/themes/livingstone/elements/container.overrides
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.ui.grid {
+
+  > .row.container.ui {
+    max-width: @maxWidth !important; /* Need to use !important to override library value */
+
+    /*
+      Since the container will handle the padding for us, no need to apply to the columns
+    */
```
marcdavi-es: We can remove this comment.

src/components/layout/Container/component.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Container as SemanticContainer } from 'semantic-ui-react';
+
+/**
+ * Container is the Lodgify UI interface for the
+ * Semantic UI Container.
+ *
```
Jonnygwi: You can remove this empty line of whitespace

src/styles/semantic/themes/livingstone/globals/site.variables
```diff
@@ -141,11 +141,12 @@
 --------------------*/
 
 @tinyBreakpoint: 400px;
-@tabletBreakpoint: 600px;
+@tabletBreakpoint: 670px;
```
marcdavi-es: Need to be careful here. We use `600` in `withResponsive` as well. This may break a bunch of existing responsive stuff.

Do we need to modify this? This is a question for Julio.

src/components/layout/Container/component.js
stephenkr: At the moment our implementation of the `Container` does not require this prop.

I'll remove it since we dont need it.

src/styles/semantic/themes/livingstone/elements/container.overrides
stephenkr: Yes its gone

src/components/layout/Container/component.js
stephenkr: Yes no problem.

src/styles/semantic/themes/livingstone/globals/site.variables
stephenkr: Yep, it is I'll ask Julio about this.

So the reasoning behind this change was to keep our breakpoints consistent, this will contribute to a cleaner user experience.

PR #319, opened 2018-09-05
TE-976 Social icon hover state

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-976)

Fix the hover state of the social media icons 

#### Before
image

#### After
image

### DIFF
--- a/src/styles/semantic/themes/livingstone/collections/menu.overrides
+++ b/src/styles/semantic/themes/livingstone/collections/menu.overrides
@@ -187,15 +187,18 @@
 ---------------*/
 
 /* Inverted */
-.ui.inverted.grey.menu {
-  background-color: @footerBottomBackground;
+.ui.inverted.menu {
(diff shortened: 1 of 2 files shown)


src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -187,15 +187,18 @@
...
-  /* Inverted Hover */
-  .item:hover {
-    background-color: @footerBottomBackground;
-    color: @footerBottomMenuColor;
+  .right.menu {
+
+    /* Inverted Hover */
+    .item:hover {
+      background-color: @footerBottomBackground;
+      color: @footerBottomMenuColor;
```
stephenkr: Since these styles are not nested under the footer, should we rename the variables to something more specific for the inverted menu rather than the footer?

PR #320, opened 2018-09-05
TE-971 Apply VerticalGutters to specified components

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-971)

Applies Vertical gutters to the following list of components
```
MEDIA

ResponsiveImage

GENERAL WIDGETS

CallMeBack
Contact
FeaturedProperties
FeaturedRoomTypes
HTML
OwnerLogin
OwnerSignUp
Promotion
Review

PROPERTY PAGE WIDGETS

Amenities
Availability
Description
HostProfile
KeyFacts
Location
PaymentInformation
Pictures
PropertySearchBar
Rates
Reviews
RoomType
Rules
SleepingArrangement
```

verticalgutters_application

amenities
availability
callmeback
contact
description
featuredproperties
featuredroomtypes
hostprofile
html
keyfacts
location
ownerlogin
ownersignup
paymentinformation
pictures
promotion
propertysearchbar
rates
responsiveimage
review
reviews
roomtype
rules
sleepingarrangements


src/components/property-page-widgets/Rates/component.js
```diff
@@ -33,57 +35,59 @@ export const Component = ({
...
-        ])}
-        tableHeadings={[
-          <Dropdown onChange={onChangeCurrency} options={currencyOptions} />,
-          ...rateHeadings,
-        ]}
-      />
-    </ShowOnDesktop>
-  </div>
+  <VerticalGutters>
+    <div>
```
marcdavi-es: We can lose this `div` now. It's only there to wrap the children.

src/components/property-page-widgets/SleepingArrangements/component.js
```diff
@@ -8,26 +8,30 @@ import { Grid } from 'layout/Grid';
...
-          mobile={4}
-          tablet={4}
-        >
-          <IconCard isLeftAligned label={label} name={iconName} />
-        </GridColumn>
-      ))}
-    </Grid>
-  </div>
+  <VerticalGutters>
+    <div>
```
marcdavi-es: I think we can lose this `div` too. Test it to make sure the interface doesn't break.

src/components/general-widgets/CallMeBack/component.js
```diff
@@ -22,6 +22,8 @@ import { TextArea } from 'inputs/TextArea';
 import { TextInput } from 'inputs/TextInput';
 import { ICON_NAMES } from 'elements/Icon';
 
+import { VerticalGutters } from '../../layout/VerticalGutters';
```
stephenkr: You can use the alias instead `layout/VerticalGutters`

src/components/general-widgets/CallMeBack/component.js
```diff
@@ -43,40 +45,42 @@ export const Component = ({
...
-      />
-      <Dropdown
-        label={propertyInputLabel}
-        name="property"
-        options={propertyOptions}
-      />
-    </InputGroup>
-    <TextArea label={notesInputLabel} name="notes" />
-  </Form>
+  <VerticalGutters>
```
stephenkr: @red-runner @marcdavi-es Just wondering, but would it be better to add an option to Grid row or grid column to add the dividers? rather than adding the divider to each component? 

We can then apply the divider from templates-ssr rather than assuming that these components will always have the divider

marcdavi-es: I don't think so.

We want to avoid writing a bunch of exceptions into `templates-ssr`

stephenkr: Oh ok, it just seems like an extra step when adding a component, it seems like there is a better way to implement this rather than having to update each component.

Such as if the Grid column had this enabled by default.

marcdavi-es: We don't want gutters on every Grid column.

We want them on some components but not others.

stephenkr: So what happens with one component with the gutters is next to another without them?

Or would this not happen due to how the component is meant to be implemented?

marcdavi-es: That's up to the user when they build their site. 

Some components are going to look bad next to other components, for a number of reasons. We can't engineer that away.

> it seems like there is a better way to implement this

I'm open to you suggesting a better alternative.

stephenkr: I was thinking about taking more of an element relationship approach, so if the component exists in a column then apply the padding.

Or if a row follows another row then apply a padding-top of `80px`, where the first child in the row list is an exception.

marcdavi-es: I thought about this. I don't think this works well considering we have deep nesting of Grids, Rows and Columns. You'll need to write in a bunch of exceptions.

While this solution is repetitive, it applies the gutters only where they are needed in an explicit way with zero exceptions. That's why I leant this way.

stephenkr: Yep true, unless the grid is aware that is the primary grid for the page and any sibling of that grid would follow the CSS rules.

I think all the css required would be:
```
.ui.grid.is-primary > .row.ui {
  + .row.ui {
    padding-top: 80px;
  }
  
  &:first-child {
    padding-top: 35px;
  }
}
```

marcdavi-es: Why is this better?

In my view it is worse because...

1. it hardwires exceptions (like `.is-primary` and `&:first-child`)
2. it relies on the implicit behaviour of recherche CSS selectors.

In any case - I don't think it works for our use case. It applies a blanket `padding-top` on all direct children of `.ui.grid.is-primary`, which is not what we need.

stephenkr: But can we assume that the components should always have the vertical spacing?
 if so then this implementation is fine.

I think the CSS approach is a cleaner approach as the code is shorter and allows the components to be implemented in other components without having a 80px vertical padding. More logic would be required to remove the padding to satisfy the requirement, hence making our code more complex.

Just addressing the two points above,
1. `is-primary` is a modifier class and the `first-child` applies a margin-top of `35px` to the first element in the grid as this is the expected behavior of the theme. What if another theme does not want this margin-top, we would have to add an if statement to be aware of the theme.
2. Most grid systems have this type of relationship, even semantic has it when a divider follows a row https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/collections/grid.less#L137 

Just my 2 cents :)

PR #323, opened 2018-09-06
TE-983 Highlight Footer content

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-983)

Enables a user to select the copyright data, phone number and property address

#### Before
image

#### After
image

### DIFF
--- a/src/styles/semantic/themes/livingstone/collections/menu.overrides
+++ b/src/styles/semantic/themes/livingstone/collections/menu.overrides
@@ -66,6 +66,11 @@
      States
 ---------------*/
 
+  /* Selectable Item */
+  .item.is-selectable {
+    user-select: auto;
+  }
+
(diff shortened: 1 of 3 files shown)


src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -66,6 +66,11 @@
      States
 ---------------*/
 
+  /* Selectable Item */
+  .item.is-selectable {
+    user-select: initial;
```
stephenkr: Be careful using `initial`, it is not supported in IE 11, so this rule will be ignored

Jonnygwi: Ahh nice!

PR #324, opened 2018-09-06
TE-1004: add size prop to set modal size in the Modal

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1004)

sets the size prop in the modal

### Reason
Because of the changes required in https://youtrack.lodgify.net/issue/TE-999, the modal is required to be 720px, this is the size small of the SemanticUi modal


src/components/elements/Modal/component.js
```diff
@@ -23,13 +23,16 @@ Component.displayName = 'Modal';
...
   isFullscreen: false,
+  size: 'tiny',
 };
 
 Component.propTypes = {
   /** The content of the modal when displayed. */
   children: PropTypes.node.isRequired,
   /** Is the modal filling the whole screen when displayed. */
   isFullscreen: PropTypes.bool,
+
```
marcdavi-es: Needs prop documentation else it's undocumented in styleguidist.

stephenkr: Oh my bad, sorry i thought i added it, adding now.

PR #325, opened 2018-09-06
TE-986 Propose theming pattern

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-986)

Proposes a pattern for implementing themable styles. Uses `a` selector as a simple example.

### Any other notes

Here's a demo of it working...

kapture 2018-09-06 at 16 27 11

Here's what's going on in my head...

#### Pros
- Does everything we need it to do.
- Straightforward code directly implementing a browser technology.
- Simple fallback to default if theming not used.
- Simple fallback to default if technology not supported.

kapture 2018-09-06 at 16 24 46

- Low commitment. If we want to change later, we aren't deeply invested in this pattern

#### Cons 
- Not supported by IE11. Note: It is likely that we will not have to fully support IE11

#### Other considerations
The main question is a CSS-in-JS vs CSS-only solution. 

I can see us adopting CSS in JS across the whole project, including theming support, once the dust has settled on CSS in JS as a practice. 

For now I lean towards maintaining the separation of concerns and including theming support in our CSS.


stephenkr:
It looks good, but do we have much traffic coming to our customer's websites that use IE11?

PR #327, opened 2018-09-07
TE-1007: amenities all

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1007)

Add prop to render extra items after default items or render extra items in modal

### Other notes
This part of the design improvements for the room type, more info can be found here https://youtrack.lodgify.net/issue/TE-999

__Preview__
screen shot 2018-09-07 at 10 58 24

### DIFF
--- a/src/components/property-page-widgets/Amenities/Readme.md
+++ b/src/components/property-page-widgets/Amenities/Readme.md
@@ -193,7 +193,7 @@ const amenities = [
 
 ```
 
-#### Display the extra items not in a modal
+#### Display the extra items in a modal
 
 ```jsx
 const amenities = [
@@ -239,6 +239,6 @@ const amenities = [
 ];
 
 
-<Amenities amenities={amenities} hasExtraItemsInModal={false} headingText="Property Amenities" />
+<Amenities hasExtraItemsInModal amenities={amenities} headingText="Property Amenities" />
 
 ```
--- a/src/components/property-page-widgets/Amenities/Readme.md
+++ b/src/components/property-page-widgets/Amenities/Readme.md
@@ -192,3 +192,53 @@ const amenities = [
 <Amenities amenities={amenities} headingText="Property Amenities" isStacked />
 
 ```
+
+#### Display the extra items not in a modal
+
+```jsx
+const amenities = [
+  {
+    name: 'Cooking',
+    iconName: 'coffee',
+    items: ['Toaster', 'Microwave', 'Coffee Machine'],
+  },
+  {
+    name: 'Bathroom',
+    iconName: 'bathroom',
+    items: ['Bidet', 'Hair Dryer'],
+  },
+  {
+    name: 'Outside',
+    iconName: 'sun',
+    items: ['Tennis Court'],
+  },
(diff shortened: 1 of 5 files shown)


src/components/property-page-widgets/Amenities/Readme.md
```diff
@@ -192,3 +192,53 @@ const amenities = [
...
+      'Steam Sauna',
+      'Fitness-Room',
+      'Parking',
+      'Laundry Service',
+    ],
+  },
+];
+
+
+<Amenities amenities={amenities} hasExtraItemsInModal={false} headingText="Property Amenities" />
```
Jonnygwi: I would rather we flipped this so that `hasExtraItemsInModal` is by default `false`. That is the pattern we've used throughout.

src/components/property-page-widgets/Amenities/Readme.md
```diff
@@ -192,3 +192,53 @@ const amenities = [
 <Amenities amenities={amenities} headingText="Property Amenities" isStacked />
 
 ```
+
+#### Display the extra items not in a modal
```
Jonnygwi: With https://github.com/lodgify/lodgify-ui/pull/327/files#r215965928 in mind. `Display the extra items not in a modal` -> `Display extra items in a modal`

src/components/property-page-widgets/Amenities/Readme.md
Jonnygwi: This would mean other components would also need to be updated

stephenkr: Yep, that makes sense, I was going to take this approach but having a backward compatible approach I felt was a better solution since fewer changes would be required.

@marcdavi-es What do you think?

src/components/property-page-widgets/Amenities/Readme.md
stephenkr: I'll implement this once a decision has been made above.

src/components/property-page-widgets/Amenities/Readme.md
marcdavi-es: @stephenkr I'd go with `false` as the default for consistency.

stephenkr: Yep no problem, i'll implement it then.

src/components/property-page-widgets/Amenities/component.js
```diff
@@ -71,6 +67,8 @@ Component.propTypes = {
       name: PropTypes.string.isRequired,
     })
   ).isRequired,
+  /** Should the extra items be in a modal. */
```
Jonnygwi: `Should the extra items be in a modal.` -> `Are the extra items displayed in a modal.`

stephenkr: Sounds good.

stephenkr:
@Jonnygwi So I can only see the `RoomType` using the `Amenities` component.

And since it doesn't need the modal, it doesn't need to be updated.

PR #329, opened 2018-09-10
TE-999: Update the RoomType modal

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-999)

Implements design improvements for the RoomType modal

### Any other notes
- Added the ClassName prop to the modal
- Rating prop has been added and is required
- The modal is `720px` to align with Semantic Ui modal sizes, agreed by Julio

__Expected design by Julio__
image1 1

__Updated modal preview__
screen shot 2018-09-10 at 10 23 48

### Other notes
- The Amenities now have a padding around them, there is another ticket to remove this https://youtrack.lodgify.net/issue/TE-1033


stephenkr:
@marcdavi-es Ok i'll take this approach over implementing CSS

stephenkr:
@marcdavi-es On point 4.

> The out the box implementation already has margin 1em between items. I don't think we need any extra CSS?

So this was added to reverse the margin, the issue is when the items wrap to the next line there is a spacing on the left side 

The css update will ensure there is no margin left.

Such as:

__With margin left__
screen shot 2018-09-12 at 09 41 52

__With margin right, the css solution__
screen shot 2018-09-12 at 09 41 45

src/components/elements/Modal/component.js
```diff
@@ -9,8 +9,15 @@ import { Icon, ICON_NAMES } from 'elements/Icon';
...
-export const Component = ({ children, isFullscreen, size, trigger }) => (
+export const Component = ({
+  className,
+  children,
+  isFullscreen,
+  size,
+  trigger,
+}) => (
   <Modal
+    className={className}
```
marcdavi-es: I don't think we need this any more?

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
```diff
@@ -30,22 +31,32 @@ export const getModalContentMarkup = (
   features,
   name,
   nightPrice,
+  ratingNumber,
   slideShowImages
 ) => (
   <Modal.Content>
     <Heading>{name}</Heading>
-    <Divider />
+    <div className="rating-container">
```
marcdavi-es: I think we can lose this `div` now?

src/styles/semantic/themes/livingstone/elements/list.variables
```diff
@@ -0,0 +1,82 @@
...
+/* Divided Horizontal */
+
+/* Divided */
+
+/* Divided Horizontal */
+
+/* Relaxed */
+
+/* Very Relaxed */
+
```
marcdavi-es: No need for this file until we write our own variables.

src/components/elements/Modal/component.js
stephenkr: True, removing.

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
stephenkr: I think the `div` is needed to keep the rating number and component together.

It's not a requirement but feels more semantic to group these elements.

src/styles/semantic/themes/livingstone/elements/list.variables
stephenkr: Oh ok, i'll remove it.

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
marcdavi-es: > It's not a requirement but feels more semantic to group these elements.

I don't understand?

stephenkr: Discussed in RL, i'll remove the container div and I have created a ticket to improve the Rating number implementation https://youtrack.lodgify.net/issue/TE-1049

PR #330, opened 2018-09-10
TE-1033: Allows the vertical spacing to be optional in `Amenities`

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1033)

Allows the vertical spacing to be optional in `Amenities`

### Other Notes
- Since Amenities may exist nested in another component, we need to remove the spacing
- Since Amenities can appear as a primary component in our pages, we need to keep the padding

__Before__
screen shot 2018-09-10 at 12 22 50

__After__
screen shot 2018-09-10 at 12 22 41

### DIFF
--- a/src/components/property-page-widgets/Amenities/Readme.md
+++ b/src/components/property-page-widgets/Amenities/Readme.md
@@ -243,7 +243,7 @@ const amenities = [
 
 ```
 
-#### Remove vertical padding
+#### Without vertical padding
 
 ```jsx
 const amenities = [
@@ -265,6 +265,6 @@ const amenities = [
 ];
 
 
-<Amenities isNested amenities={amenities} headingText="Property Amenities" />
+<Amenities hasVerticalGutters={false} amenities={amenities} headingText="Property Amenities" />
 
 ```
--- a/src/components/property-page-widgets/Amenities/Readme.md
+++ b/src/components/property-page-widgets/Amenities/Readme.md
@@ -242,3 +242,29 @@ const amenities = [
 <Amenities hasExtraItemsInModal amenities={amenities} headingText="Property Amenities" />
 
 ```
+
+#### Remove vertical padding
+
+```jsx
+const amenities = [
+  {
+    name: 'Cooking',
+    iconName: 'coffee',
+    items: ['Toaster', 'Microwave', 'Coffee Machine'],
+  },
+  {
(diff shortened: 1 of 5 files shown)


src/components/property-page-widgets/Amenities/Readme.md
```diff
@@ -242,3 +242,29 @@ const amenities = [
...
+  },
+  {
+    name: 'Outside',
+    iconName: 'sun',
+    items: ['Tennis Court'],
+  }
+];
+
+
+<Amenities isNested amenities={amenities} headingText="Property Amenities" />
```
marcdavi-es: The prop name `isNested` couples this use case to the component API. 

There are a bunch of reasons why you might not want the vertical gutters, so let's name the prop according to what it does.

I suggested `hasVerticalGutters`, defaulting to `true`

src/components/property-page-widgets/Amenities/Readme.md
```diff
@@ -242,3 +242,29 @@ const amenities = [
 <Amenities hasExtraItemsInModal amenities={amenities} headingText="Property Amenities" />
 
 ```
+
+#### Remove vertical padding
```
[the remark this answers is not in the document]
stephenkr: Good idea, i'll change this.

src/components/property-page-widgets/Amenities/Readme.md
stephenkr: Ah good idea, i'll change the prop name

src/components/property-page-widgets/Amenities/utils/getAmenityMarkup.js
```diff
@@ -0,0 +1,43 @@
...
+import { getCategoryMarkup } from './getCategoryMarkup';
+import { getExtraItemsMarkup } from './getExtraItemsMarkup';
+
+/**
+ * @param {Object[]}  amenities
+ * @param {boolean}   hasExtraItemsInModal
+ * @param {string}    headingText
+ * @param {boolean}   isStacked
+ * @param {string}    modalTriggerText
+ */
```
marcdavi-es: Please add a `@returns` tag.

stephenkr: Fixing now.

PR #331, opened 2018-09-10
TE-1034: active link inherit background color for inverted `Menu`

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1034)

Fixes active link style for inverted `Menu`

__Before__
screen shot 2018-09-10 at 12 43 04

__After__
screen shot 2018-09-10 at 12 42 57

### DIFF
--- a/src/styles/semantic/themes/livingstone/collections/menu.overrides
+++ b/src/styles/semantic/themes/livingstone/collections/menu.overrides
@@ -295,7 +295,7 @@ header > .ui.menu.text {
 footer {
 
   .top-navigation {
-    background-color: @black;
+    background-color: @footerTopBackground;
   }
 
   .bottom-navigation {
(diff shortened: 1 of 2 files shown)


src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -197,6 +197,10 @@
 
   /* Inverted Sub Menu */
 
+  .link.item:active {
+    background-color: inherit;
+  }
```
marcdavi-es: Rather than adding more css, we can just modify the `@invertedHoverBackground` variable.

stephenkr: Yep good point, i'll give it a try

stephenkr: On review, its not the hover state we need to modify, its the active state, i'll see if i can take a variable first approach.

PR #333, opened 2018-09-10
TE-988 Nested dropdown options

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-988)

Adds ability to indent dropdown options, unit tests and adds an example.

### Any other notes
- Added the accepted values in documentation because nested shapes don't get the "One of: " documentation.

image

### DIFF
--- a/src/components/inputs/Dropdown/component.js
+++ b/src/components/inputs/Dropdown/component.js
@@ -118,6 +118,8 @@ Component.propTypes = {
     PropTypes.shape({
       /** The source url for the image to display with the option. */
       image: PropTypes.string,
+      /** The indent level of an option. One of: 1, 2 */
+      indent: PropTypes.oneOf([1, 2]),
       /** The visible text for the option. */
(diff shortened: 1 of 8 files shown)


src/components/inputs/Dropdown/component.js
```diff
@@ -118,6 +118,8 @@ Component.propTypes = {
     PropTypes.shape({
       /** The source url for the image to display with the option. */
       image: PropTypes.string,
+      /** The indent level of an option. One of: 1, 2 */
```
marcdavi-es: No need for `One of: 1, 2` ... that will show up in the docs parsed from the `PropTypes` entry

Jonnygwi: @marc nested props in a shapes generate those docs unfortunately.

marcdavi-es: o m g

PR #335, opened 2018-09-12
TE-1035: icon should be the correct width in Button

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1035)

updates icon width in Button

__Before__
screen shot 2018-09-12 at 13 06 40

__After__
screen shot 2018-09-12 at 13 06 24


stephenkr:
@red-runner yep that's a good point, i'll do some investigation to see if i can improve the vertical alignment, although the alignment is consistent with the original version

PR #336, opened 2018-09-12
TE-987 Description component fixes

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-987)

Improves `Description` component icon disposition in mobile and desktop.

### Any other notes

description_refactor


src/components/property-page-widgets/Description/component.spec.js
```diff
@@ -33,13 +37,28 @@ const props = {
...
+    size: null,
+  },
   propertyMainCharacteristics,
   propertyName: 'Yolo',
   propertyType: 'Bed & Breakfast',
 };
 
 const getDescription = extraProps =>
   shallow(<Description {...props} {...extraProps} />);
+const getDesktop = getDescription().find(ShowOnDesktop);
```
stephenkr: Can you make this into a function, just so you get a fresh copy for each spec?

Also, I'd call this `getDesktopMarkup`, same for `getShowOnMobile ` to `getShowOnMobileMarkup`

src/components/property-page-widgets/Description/component.spec.js
```diff
@@ -231,16 +319,26 @@ describe('<Description />', () => {
...
-  describe('each `Icon`', () => {
-    it('shoudl have the right props', () => {
+  describe('the first `Icon`', () => {
+    it('should have the right props', () => {
       const wrapper = getDescription()
+        .find(GridColumn)
+        .at(8)
         .find(Icon)
-        .at(4);
+        .at(0);
```
stephenkr: You can simplify this by just selecting the nth Icon, such as

```
getDescription().find(Icon).at(nth)
```
where nth is the index of the Icon

src/components/property-page-widgets/Description/component.spec.js
```diff
@@ -98,23 +117,92 @@ describe('<Description />', () => {
...
+        wrapper,
+        ...getArrayOfLengthOfItem(4, GridColumn)
+      );
+    });
+
+    describe('the `Icon` nested in the first `GridColumn`', () => {
+      const wrapper = getShowOnMobile
+        .find(GridColumn)
+        .first()
+        .children();
```
stephenkr: When i see `.children` i think there are multiple elements, could you select the element directly or just use `firstChild` instead?

src/components/property-page-widgets/Description/component.spec.js
```diff
@@ -98,23 +117,92 @@ describe('<Description />', () => {
...
+        wrapper,
+        ...getArrayOfLengthOfItem(4, ListItem)
+      );
+    });
+
+    describe('the `Icon` nested in the first `ListItem`', () => {
+      const wrapper = getDesktop
+        .find(ListItem)
+        .first()
+        .children();
```
stephenkr: Related to comment https://github.com/lodgify/lodgify-ui/pull/336/files#r217057677

src/components/property-page-widgets/Description/component.spec.js
```diff
@@ -98,23 +117,92 @@ describe('<Description />', () => {
     });
   });
 
+  describe('the first `Heading`', () => {
+    it('should have the right props', () => {
+      const wrapper = getDescription().find(Heading);
```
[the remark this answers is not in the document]
red-runner: @stephenkr This component just has one Heading. It is described as _first heading_ because there is a sub heading (perhaps that part is unclear).  I've checked in other tests and it seems that we haven't been as precise. 

Could we leave it on synthesising this part? I think .first() isn't necessary in this case.

```const getHeading = () => getDescription().find(Heading)```

File: src/components/property-page-widgets/Description/component.spec.js (hunk 2 above)
Comment (red-runner) [reply src/components/property-page-widgets/Description/component.spec.js]: I believe I already went with this approach

```
  describe('each of the propertyKeyFact `Icon`s', () => {
    it('should have the right props', () => {
      const wrapper = getDescription()
        .find(Icon)
        .at(0); // where 0 is the index of the Icon
```

File: src/components/property-page-widgets/Description/component.spec.js (hunk 3 above)
Comment (red-runner) [reply src/components/property-page-widgets/Description/component.spec.js]: I've search through the project for any instance of firstChild and I couldn't find any. Instead I've went for the following approach, just to be consistent:

```
    describe('the `Icon` nested in the first `GridColumn`', () => {
      const wrapper = getMobileMarkup.find(Icon).at(0);
```

PR #337, opened 2018-09-12
TE-1047: Add theme variables for Button

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1047)

Adds CSS variables for theming the `Button`

### Any other notes
- The variable names contain `action` as opposed to `button` to align with the `palette` object in the API
- I assume that the text color and hover color will be computed from the button background color

__Preview__
screen shot 2018-09-12 at 15 59 20


src/styles/constants.js
```diff
@@ -1,6 +1,9 @@
 const THEME_CUSTOM_PROPERTY_PREFIX = '--lodgify-ui';
 
 export const THEME_CUSTOM_PROPERTY_IDENTIFIERS = {
+  ACTION_BACKGROUND_COLOR: `${THEME_CUSTOM_PROPERTY_PREFIX}-action-background-color`,
+  ACTION_HOVER_BACKGROUND_COLOR: `${THEME_CUSTOM_PROPERTY_PREFIX}-action-hover-background-color`,
+  ACTION_TEXT_COLOR: `${THEME_CUSTOM_PROPERTY_PREFIX}-action-text-color`,
```
marcdavi-es: I would decouple the name from the use case. `ACTION_BACKGROUND_COLOR` will be used for `background-color` for a default button but `color` for simple button (which is what we use for the non-href `Link` component).

Suggested alternatives...

```js
ACTION_COLOR
ACTION_DARKER_COLOR
ACTION_CONTRAST_COLOR
```

Note that this invalidates what I did for `LINK_*`. Please also change `LINK_HOVER_COLOR` to `LINK_DARKER_COLOR`

marcdavi-es:
1. Suggested change inline. 
1. Don't forget to add any identifiers to `THEME_CUSTOM_PROPERTY_IDENTIFIERS` in `src/styles/constants.js`

src/styles/constants.js
stephenkr: I was more thinking to align with the CSS rule names, but this approach above would provide a more flexible implementation.

src/styles/semantic/themes/livingstone/elements/button.overrides
```diff
@@ -51,11 +53,20 @@
...
+}
+
 .ui.button.secondary:hover {
   transform: @secondaryTransformHover;
 }
 
 /* Focused */
+.ui.button:focus {
+  background-color: ~"var(@{themeActionDarkerColorIdentifier}, @{themeActionDarkerColorDefault});";
+  color: ~"var(@{themeActionColorIdentifier}, @{themeActionColorDefault});";
```
marcdavi-es: For each of these, we can use the straightforward variable interpolation syntax like https://github.com/lodgify/lodgify-ui/pull/337/files#diff-9db5ca13d272f5d2fb6018e39a280aa5R95

src/styles/semantic/themes/livingstone/elements/button.variables
```diff
@@ -46,6 +50,8 @@
 
 /* Hovered */
 @hoverBackgroundColor: @darkPrimaryColor;
+@themeActionDarkerColorIdentifier: ~"@{themeCustomPropertyPrefix}-action-darker-color";
+@themeActionDarkerColorDefault: @darkPrimaryColor;
```
marcdavi-es: I _think_ all these variable definitions belong up in `site.variables`. They will be consumed by other components too.

PR #344, opened 2018-09-14
TE-1031 HomepageHero

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1031)

Exposes, unit tests and adds examples for the HomepageHero component

### Any other notes

screen shot 2018-09-14 at 15 09 18

image


src/styles/semantic/themes/livingstone/elements/segment.overrides
```diff
@@ -220,6 +220,12 @@
     flex: 1;
   }
 
+  .ui.container:not(.menu) {
```
stephenkr: Just thinking but this would be better in the `container.overrides`

src/components/collections/HomepageHero/component.js
```diff
@@ -0,0 +1,191 @@
...
+  isSearchBarDisplayedAsModal: false,
+  isSearchBarShowingLocationDropdown: false,
+  isSearchBarShowingSummary: false,
+  searchBarOnSubmit: Function.prototype,
+  searchBarModalHeadingText: CHECK_OUR_AVAILABILITY,
+  searchBarModalSummaryElement: null,
+  searchBarModalTrigger: <Icon name={ICON_NAMES.SEARCH} />,
+  searchBarSearchButton: (
+    <Button icon={ICON_NAMES.SEARCH} isPositionedRight isRounded>
+      Search
```
stephenkr: This should be extracted out to i18n keys

src/components/collections/HomepageHero/component.js
```diff
@@ -0,0 +1,191 @@
...
+
+Component.propTypes = {
+  /** The background image url of the hero. */
+  backgroundImageUrl: PropTypes.string.isRequired,
+  /**
+   * A function called for each day to be displayed in the DateRangePicker. Returning true blocks that day in the date range picker.
+   * @param   {Moment}  day - The day to test.
+   * @returns {boolean}     - Is the day blocked.
+   */
+  getSearchBarIsDayBlocked: PropTypes.func,
```
stephenkr: What about `searchBarGetIsDayBlocked`, just to prefix `getIsDayBlocked`, looks strange when reading it

src/components/collections/HomepageHero/component.js
```diff
@@ -0,0 +1,191 @@
...
+    </Button>
+  ),
+  searchBarSummaryElement: null,
+};
+
+Component.propTypes = {
+  /** The background image url of the hero. */
+  backgroundImageUrl: PropTypes.string.isRequired,
+  /**
+   * A function called for each day to be displayed in the DateRangePicker. Returning true blocks that day in the date range picker.
```
stephenkr: Small thing, but move the `Returning true bloc....` to a new line in the comment

src/components/collections/HomepageHero/component.js
```diff
@@ -0,0 +1,191 @@
...
+  searchBarLocationOptions,
+  searchBarModalHeadingText,
+  searchBarModalSummaryElement,
+  searchBarModalTrigger,
+  searchBarOnSubmit,
+  searchBarSearchButton,
+  searchBarSummaryElement,
+}) =>
+  isUserOnMobile ? (
+    <Hero
```
stephenkr: I think this could be cleaner, what about having a default props object and having one for mobile, such as:

```
const props = {
  backgroundImageUrl: backgroundImageUrl,
  headerLogoSrc: headerLogoSrc,
  headerLogoText: headerLogoText,
  headerNavigationItems: headerNavigationItems,
  headerPrimaryCTA: headerPrimaryCTA,
  headerSearchBarGuestsOptions: searchBarGuestsOptions,
  headerSearchBarLocationOptions: searchBarLocationOptions,
  heading: heading,
}

const mobileProps = {
  ...props,
  extraContent: getSearchBarMarkup(
    getSearchBarIsDayBlocked,
    isSearchBarDisplayedAsModal,
    isSearchBarShowingLocationDropdown,
    isSearchBarShowingSummary,
    searchBarGuestsOptions,
    searchBarLocationOptions,
    searchBarModalHeadingText,
    searchBarModalSummaryElement,
    searchBarModalTrigger,
    searchBarOnSubmit,
    searchBarSearchButton,
    searchBarSummaryElement
  )
}

return isUserOnMobile ? (
  <Hero {...props} />
) :  (
  <Hero {...mobileProps} />
);
```

src/components/collections/HomepageHero/utils/getSearchBarMarkup.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param  {string}   modalHeadingText
+ * @param  {Node}     modalSummaryElement
+ * @param  {Node}     modalTrigger
+ * @param  {Function} onSubmit
+ * @param  {Node}     searchButton
+ * @param  {Node}     summaryElement
+ * @return {Object}
+ */
+export const getSearchBarMarkup = (
+  getIsDayBlocked,
```
stephenkr: Might be cleaner to just pass this as an object, and then for the `SearchBar` just do:

```
<SearchBar {...props} />
```

src/styles/semantic/themes/livingstone/elements/segment.overrides
Jonnygwi: This rule is only applied when it's in a `.ui.segment.is-hero` so I think it belongs in this file. thoughts?

stephenkr: Oh interesting, although it doesn't directly change the styling of `Segment`, it changes the styling of the container.

And in the `container.overrides` you'll need to wrap this style with a segment.

I think it makes more sense in the `container.overrides` since it keeps container related styling in one place.

src/components/collections/HomepageHero/Readme.md
```diff
@@ -0,0 +1,55 @@
...
+    headerNavigationItems={navigationItems}
+    headerPrimaryCTA={{ href: '/book', text: 'Book now'}}
+    heading="Super Interesting Heading"
+    searchBarGuestsOptions={guestsOptions}
+    searchBarLocationOptions={locationOptions}
+    isSearchBarShowingLocationDropdown
+  />
+```
+
+### With property summary
```
marcdavi-es: I don't think this use case is intended?

We can remove this example.

src/components/collections/HomepageHero/component.js
```diff
@@ -0,0 +1,189 @@
...
+import { CHECK_OUR_AVAILABILITY } from 'utils/default-strings';
+import { Icon, ICON_NAMES } from 'elements/Icon';
+import { Button } from 'elements/Button';
+import { SEARCH } from 'utils/default-strings';
+
+import { getSearchBarMarkup } from './utils/getSearchBarMarkup';
+
+/**
+ * A homepage hero displays a header, heading and search bar
+ * in a full width and hight container
```
marcdavi-es: 1. `height`
2. Can we be clearer here on what 'full' is relative to? I think `in a container which fills the full width and height of the viewport` is clearer.

PR #345, opened 2018-09-14
TE-1066: Add in reduced version for `Summary`

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1066)

Adds in a reduced version of the `Summary` component

__Before__
screen shot 2018-09-14 at 15 16 09

__After__
screen shot 2018-09-14 at 15 16 15

### DIFF
--- a/src/components/property-page-widgets/Summary/component.js
+++ b/src/components/property-page-widgets/Summary/component.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Rating, Segment } from 'semantic-ui-react';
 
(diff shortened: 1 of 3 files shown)


src/components/property-page-widgets/Summary/component.js
```diff
@@ -11,37 +11,61 @@ import { getNightPriceMarkup } from 'utils/get-night-price-markup/';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export const Component = ({
+  hasReducedVersion,
```
Jonnygwi: I guess even if we are using this naming this should be something more like `isReducedVersion`

PR #347, opened 2018-09-18
TE-1046 Header and SearchBar LocationOptions

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1046)

Removes `props.isShowingLocationDropdown` from SearchBar as we can just check if there are any being passed into the component and updates the Header to correctly render the SearchBar depending on the options available

### Any other notes
kapture 2018-09-18 at 16 32 03

### DIFF
--- a/src/components/general-widgets/SearchBar/utils/getFormFieldMarkup.js
+++ b/src/components/general-widgets/SearchBar/utils/getFormFieldMarkup.js
@@ -1,6 +1,7 @@
 /* eslint react/prop-types: 0 */
 import React, { Fragment } from 'react';
 import { Form } from 'semantic-ui-react';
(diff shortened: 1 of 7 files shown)


src/components/general-widgets/SearchBar/utils/getFormFieldMarkup.js
```diff
@@ -46,7 +45,7 @@ export const getFormFieldMarkup = (
           />
         </Form.Field>
       )}
-      {!!isShowingLocationDropdown && (
+      {!!size(locationOptions) > 0 && (
```
stephenkr: You can remove the `!!` prefixing `size`

PR #351, opened 2018-09-19
TE-1080 HomepageHero search bar dropdowns

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1080)

Changes the HomepageHero dropdowns to always open above. 

### Any other notes
image

### DIFF
--- a/src/components/general-widgets/SearchBar/Readme.md
+++ b/src/components/general-widgets/SearchBar/Readme.md
@@ -40,6 +40,16 @@ const { guestsOptions, locationOptions } = require('./mock-data/options');
 <SearchBar
   guestsOptions={guestsOptions}
 />
+
+```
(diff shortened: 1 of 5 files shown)


src/components/general-widgets/SearchBar/Readme.md
```diff
@@ -40,6 +40,16 @@ const { guestsOptions, locationOptions } = require('./mock-data/options');
...
 />
+
+```
+#### Dropdowns open above
+```jsx
+const { guestsOptions, locationOptions } = require('./mock-data/options');
+
+<SearchBar
+  guestsOptions={guestsOptions}
+  willDropdownsOpenAbove={true}
```
stephenkr: No need for the `={true}`, just having the prop should be enough

PR #352, opened 2018-09-21
TE-1091 Fix the max height and width of header logo

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1091)

Adds a max width and height to the header's logo

#### Before
image

#### After
image


marcdavi-es:
I preferred the way it was before.

PR #361, opened 2018-09-25
TE-1100 Gallery: expose component with tests and examples

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1100)

Exposes `Gallery` component with tests and examples

#### Mobile

screen shot 2018-09-26 at 11 02 21

#### Desktop

screen shot 2018-09-26 at 11 02 06

### DIFF
--- a/src/components/media/Gallery/component.js
+++ b/src/components/media/Gallery/component.js
@@ -1,0 +1,79 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+import { Modal } from 'elements/Modal';
+import { HorizontalGutters } from 'layout/HorizontalGutters';
+import { Divider } from 'elements/Divider';
+import { Grid } from 'layout/Grid';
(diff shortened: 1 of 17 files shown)


src/components/media/Gallery/component.js
```diff
@@ -0,0 +1,77 @@
...
+      {heading && (
+        <Fragment>
+          {heading}
+          <Divider hasLine />
+        </Fragment>
+      )}
+      <Grid columns={2} stackable>
+        <GridRow>
+          {images.map(({ label, ...otherImageProps }, index) => (
+            <GridColumn key={buildKeyFromStrings(label, index)} width={6}>
```
stephenkr: For the `columns` prop in the Grid to be respected, you should now set the `width` in the GridColumn to null

marcdavi-es: Aah. I see. 

OK. Would it be better to pass no `width` prop? Rather than `null`?

PR #363, opened 2018-09-26
TE-1081: support no padding for the GridColumn

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1081)

Adds the option to remove the GridColumn padding

### Any other notes
- The use case is for the root GridColumn in `TemplatesSSR`
-- we wrap the rows with a GridColumn to prevent the rows from being inline
-- the issue is a padding of `1rem` is applied to the top and bottom of the page
-- this update will remove this.

- This change is under `TE-1081` since it affects the work on VerticalPadding in `TemplatesSSR`

### DIFF
--- a/src/components/layout/GridColumn/component.js
+++ b/src/components/layout/GridColumn/component.js
@@ -1,24 +1,34 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Grid } from 'semantic-ui-react';
+import getClassNames from 'classnames';
 
 /**
  * GridColumn is the Lodgify UI interface for the
  * Semantic UI Grid.Column.
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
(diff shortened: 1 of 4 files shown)


src/components/layout/GridColumn/component.js
```diff
@@ -1,24 +1,34 @@
...
 Component.displayName = 'GridColumn';
 
 Component.defaultProps = {
+  hasNoPadding: false,
   verticalAlignContent: 'top',
   width: 12,
 };
 
 Component.propTypes = {
+  /** Ensure the column has no padding */
```
marcdavi-es: `/** Does the column have no padding */ `

stephenkr: Ah very good, updating now.

stephenkr:
From a bit more investigation in the Grid in templatesSSR, i dont think this change is necessary anymore. I'll close it.

PR #369, opened 2018-09-27
TE-1088: improve chevron positioning in dropdown

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1088)

Improves the positioning of the chevron in the dropdown

### Any other notes
- This update not only affects the header menu dropdown, but also the footer upward dropdown and the submenu dropdown
- ~If a SemanticUI variable existed for this I would have used it, the issue is the left positioning of 50% is hard coded. https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/dropdown.less#L1253~

### Previews

__Before__
screen shot 2018-09-27 at 09 36 53
screen shot 2018-09-27 at 09 36 20
screen shot 2018-09-27 at 09 35 24

__After__
screen shot 2018-09-27 at 09 36 47
screen shot 2018-09-27 at 09 36 02
screen shot 2018-09-27 at 09 35 04

### DIFF
--- a/src/styles/semantic/themes/livingstone/modules/dropdown.overrides
+++ b/src/styles/semantic/themes/livingstone/modules/dropdown.overrides
@@ -341,6 +341,15 @@
       color: @textColor;
     }
   }
+
+  &.top.right.ui,
+  &.top.left.ui {
+    margin-left: @floatedDistance !important;
+
+    .menu & > .menu {
+      margin-top: @pointingDropdownMenuDistance;
+    }
+  }
 }
 
 /* Pointing Upward */
--- a/src/styles/semantic/themes/livingstone/modules/dropdown.overrides
+++ b/src/styles/semantic/themes/livingstone/modules/dropdown.overrides
@@ -341,15 +341,6 @@
       color: @textColor;
     }
   }
-
-  &.top.right.ui,
-  &.top.left.ui {
-    margin-left: @floatedDistance !important;
-
-    .menu & > .menu {
-      margin-top: @pointingDropdownMenuDistance;
-    }
-  }
 }
 
 /* Pointing Upward */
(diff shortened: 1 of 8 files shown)


stephenkr:
@marcdavi-es I'll investigate into it, but it looks like the chevron position is hardcoded as seen here https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/dropdown.less#L1253

marcdavi-es:
@stephenkr That's the default.

You can see from line 1257 of the same file to the end of the file the styles for all of the alternative chevron positions.

These are applied according to the `pointing` prop passed to `Dropdown`

stephenkr:
On the latest change.

- Applied the positioning prop `top left` to ensure SemanticUI places the chevron in the top left of the dropdown menu
- Improved the positioning of the chevron if the menu item is the last item in a float right menu (Such as in the `Header`), the chevron should appear in the top right if this is the case since the menu item content will expand to the left 
- Improved the menu position for when the position prop is set to `top left`. The menu appeared after the menu item and not in the expected position

src/styles/semantic/themes/livingstone/modules/dropdown.overrides
```diff
@@ -341,6 +341,15 @@
...
   }
+
+  &.top.right.ui,
+  &.top.left.ui {
+    margin-left: @floatedDistance !important;
+
+    .menu & > .menu {
+      margin-top: @pointingDropdownMenuDistance;
+    }
+  }
```
stephenkr: The CSS is still required to have this dropdown appear as expected, applying the `pointing` prop will apply the following changes to the dropdown:
screen shot 2018-10-01 at 12 50 36

After the above changes, this dropdown will appear as:
screen shot 2018-10-01 at 12 53 18

PR #375, opened 2018-10-04
TE-1152: In Hero the Header dropdown should be above the content

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1152)

Ensure the Header is above the content in the `Hero`

__Before__
kapture 2018-10-04 at 15 34 48

__After__
after

### DIFF
--- a/src/styles/semantic/themes/livingstone/elements/segment.overrides
+++ b/src/styles/semantic/themes/livingstone/elements/segment.overrides
@@ -91,7 +91,11 @@
 
     > * {
       position: relative;
-      z-index: 1;
+      z-index: @heroContentZIndex;
+    }
+
+    > header {
+      z-index: @heroContentZIndex + 1;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/styles/semantic/themes/livingstone/elements/segment.overrides
```diff
@@ -91,7 +91,11 @@
 
     > * {
       position: relative;
-      z-index: 1;
+      z-index: @heroContentZIndex;
+    }
+
+    > header {
+      z-index: @heroContentZIndex + 1;
```
red-runner: @stephenkr why are doing `@heroContentZIndex + 1` instead of setting `@heroContentZIndex: 2` directly

stephenkr: So the above selector to select all items in the `Hero` has a `z-index` of 1, here we are just incrementing that value for the `Header`

red-runner: True! I missed that. Thanks 🎉

PR #387, opened 2018-10-10
TE-1171: use the responsive wrapper when rendering thumbnails in Pictures

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1171)

Refactor the `Pictures` component to use the `withResponsive` wrapper

### Any other notes
This update should fix an issue we are having in `templatesSSR`
- When the page is rendered we are getting the mobile version of the component instead of the desktop version.
- From looking into the SemanticUI `Responsive` component, I can see the correct component being returned when the desktop version is to be shown.
- But React does not render the desktop version, instead, it keeps the mobile version in the DOM.
- So i think using the `withResponsive` to replace the attributes in the `Thumbnail` should trigger the DOM to update these components.
- I have tested this as much as I could using TemplatesSSR, but I won't be sure until this has been published.

__Before__
We get the mobile version of `Pictures` when the desktop version is rendered.

__After__
We get the correct version depending on the viewport size

### DIFF
--- a/src/components/property-page-widgets/Pictures/component.js
+++ b/src/components/property-page-widgets/Pictures/component.js
@@ -1,14 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { withResponsive } from 'utils/with-responsive';
 import { EXPLORE_ALL_PICTURES, PROPERTY_PICTURES } from 'utils/default-strings';
 import { getFirstSixItems } from 'utils/get-first-six-items';
 import { buildKeyFromStrings } from 'utils/build-key-from-strings';
 import { Grid } from 'layout/Grid';
 import { GridRow } from 'layout/GridRow';
 import { GridColumn } from 'layout/GridColumn';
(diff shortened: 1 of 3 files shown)


marcdavi-es:
Let's talk about this guy. I'm investigating a similar bug for [TE-1187](https://youtrack.lodgify.net/issue/TE-1187)

stephenkr:
@marcdavi-es I think the issue is because of how the Hydrate method works, mentioned here https://reactjs.org/docs/react-dom.html#hydrate

> There are no guarantees that attribute differences will be patched up in case of mismatches. This is important for performance reasons because in most apps, mismatches are rare, and so validating all markup would be prohibitively expensive.

marcdavi-es:
I've looked at a number of ways around this and I think that even in the best case scenario, Semantic UI `Responsive` is unreliable in combination with SSR...

bad

I suggest we refactor `withResponsive`, `ShowOnDesktop` and `ShowOnMobile` to use css media queries instead.

stephenkr:
Closing as this conflicts with the chosen solution

PR #388, opened 2018-10-11
TE-1197 House rules ui fixes

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1197)

Fixes the UI so that the correct icons are present and collapses the rules column if none are present.

### Any other notes
image


src/components/property-page-widgets/Rules/component.js
```diff
@@ -28,22 +30,24 @@ export const Component = ({
...
       <Heading>{headingText}</Heading>
     </GridColumn>
-    <GridColumn computer={3} tablet={5}>
-      <List
-        items={rules.map(rule => (
-          <Paragraph>{rule}</Paragraph>
-        ))}
-      />
-    </GridColumn>
+    {!!size(rules) && (
```
stephenkr: We should try to keep consistent with this comparison, ideally `size(rules) > 0`

PR #389, opened 2018-10-11
TE-1170: Move the close icon in the Gallery

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1170)

Places the close icon on the right side of the modal in the `Gallery` to align with the horizontal gutter

### Other info
- Extracted out the horizontal gutter max width since it is needed in the Modal styles
- Discussed with Julio, if `isFullscreen` is true, we can assume the `HorizontalGutters` are being applied.
- The mobile version should keep the icon in the same position, the top right
- This has been tested with the  `Header` and `SearchForm` fullscreen modals

### Before
screen shot 2018-10-11 at 14 34 00
screen shot 2018-10-11 at 14 33 38

### After
kapture 2018-10-15 at 12 43 17

### DIFF
--- a/src/components/media/Gallery/component.js
+++ b/src/components/media/Gallery/component.js
@@ -1,5 +1,6 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
+import getClassNames from 'classnames';
 
 import { Modal } from 'elements/Modal';
 import { HorizontalGutters } from 'layout/HorizontalGutters';
@@ -16,7 +17,13 @@ import { ResponsiveImage } from 'media/ResponsiveImage';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export const Component = ({ heading, images, trigger }) => (
-  <Modal isFullscreen trigger={trigger}>
+  <Modal
+    className={getClassNames('with-horizontal-gutters', {
+      'with-heading': !!heading,
+    })}
+    isFullscreen
+    trigger={trigger}
+  >
     <HorizontalGutters>
       <Divider />
       {heading && (
@@ -28,7 +35,10 @@ export const Component = ({ heading, images, trigger }) => (
       <Grid columns={2} stackable>
         <GridRow>
           {images.map(({ label, ...otherImageProps }, index) => (
-            <GridColumn key={buildKeyFromStrings(label, index)}>
+            <GridColumn
+              className={`grid-column-${index}`}
+              key={buildKeyFromStrings(label, index)}
+            >
               <Heading size="small">{label}</Heading>
               <ResponsiveImage {...otherImageProps} />
               <Divider />
--- a/src/components/media/Gallery/component.js
+++ b/src/components/media/Gallery/component.js
@@ -1,6 +1,5 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
-import getClassNames from 'classnames';
 
 import { Modal } from 'elements/Modal';
 import { HorizontalGutters } from 'layout/HorizontalGutters';
@@ -17,13 +16,7 @@ import { ResponsiveImage } from 'media/ResponsiveImage';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export const Component = ({ heading, images, trigger }) => (
-  <Modal
-    className={getClassNames('with-horizontal-gutters', {
-      'with-heading': !!heading,
-    })}
-    isFullscreen
-    trigger={trigger}
-  >
+  <Modal isFullscreen trigger={trigger}>
     <HorizontalGutters>
       <Divider />
       {heading && (
--- a/src/styles/semantic/themes/livingstone/globals/site.variables
+++ b/src/styles/semantic/themes/livingstone/globals/site.variables
@@ -429,3 +429,11 @@
 @searchBarFixedZIndex: 999;
 @searchBarFixedPadding: @20px 0;
 @searchBarFixedBoxShadow: 0 -(@4px) @9px -(@6px) rgba(0, 0, 0, 0.29);
+
+/*-------------------
+   Horizontal Gutter
+--------------------*/
+
+/* Minimum Gutter is used to determine  the maximum container width for a given device */
+
+@horizontalGutterMaxWidth: 1120px;
--- a/src/styles/semantic/themes/livingstone/globals/site.variables
+++ b/src/styles/semantic/themes/livingstone/globals/site.variables
@@ -434,6 +434,8 @@
    Horizontal Gutter
 --------------------*/
 
-/* Minimum Gutter is used to determine  the maximum container width for a given device */
+/* Minimum Gutter is used to determine the maximum container width for a given device */
 
 @horizontalGutterMaxWidth: 1120px;
+// Must be defined in pixels for accuracy
+@horizontalGutterLeftRightPadding: 40px;
--- a/src/styles/semantic/themes/livingstone/modules/modal.overrides
+++ b/src/styles/semantic/themes/livingstone/modules/modal.overrides
@@ -65,14 +65,14 @@
     }
 
     .ui.divider:first-child + .grid {
-      @media (max-width: (@tabletBreakpoint - 1)) {
+      @media only screen and (max-width: (@tabletBreakpoint - 1)) {
 
         .grid-column-0 .ui.header {
           width: @reducedHeadingWidth;
         }
       }
 
-      @media (min-width: @tabletBreakpoint) {
+      @media only screen and (min-width: @tabletBreakpoint) {
 
         .grid-column-1 .ui.header {
           width: @reducedHeadingWidth;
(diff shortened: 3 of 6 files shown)


src/components/elements/Modal/component.js
```diff
@@ -22,13 +29,16 @@ export const Component = ({ children, isFullscreen, size, trigger }) => (
...
+  className: '',
   isFullscreen: false,
   size: 'tiny',
 };
 
 Component.propTypes = {
   /** The content of the modal when displayed. */
   children: PropTypes.node.isRequired,
+  /** Custom class name string to customize the modal. */
+  className: PropTypes.string,
```
marcdavi-es: The standard `lodgify-ui` pattern for modifying components is to expose specific props, usually as booleans (e.g. `isFullscreen`). We can consume these props to add styles conditionally.

The decision to expose totally open props like `className` is library-wide and we shouldn't arbitrarily introduce it in individual components.

Please refactor this out.

src/components/media/Gallery/component.js
```diff
@@ -16,7 +17,13 @@ import { ResponsiveImage } from 'media/ResponsiveImage';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export const Component = ({ heading, images, trigger }) => (
-  <Modal isFullscreen trigger={trigger}>
+  <Modal
+    className={getClassNames('with-horizontal-gutters', {
+      'with-heading': !!heading,
+    })}
```
marcdavi-es: This logic belongs in `Modal`

marcdavi-es:
> Added support for a `with-horizontal-gutters` class for the modal, in case we decide to position the close icon in the same way in another modal

- I'm going to check with @jams676 but I assume all full screen modals should have the content wrapped in `HorizontalGutters`. If that is true, we can use the existing `.fullscreen` class and avoid adding more styles.

> Added support for a `with-heading` class for the modal

- I can't understand what this achieves. Can you please describe this in real terms, not in code terms.

src/components/elements/Modal/component.js
stephenkr: Ok i'll refactor the implementation of className to be a specific prop.

stephenkr:
Hi @marcdavi-es 

> I can't understand what this achieves. Can you please describe this in real terms, not in code terms.

Applying the `with-header` class is used to complement the implementation of the `with-horizontal-gutters` class, we need a way to understand if the modal has a main heading or not.

When talking with @jams676, he requested for the heading closest to the close icon to have a reduced width to prevent any content overlapping with the icon. Having awareness around the main heading's existence would change the implementation of how these selectors are to be defined.

Adding the class also greatly reduced the complexity of the selectors when applying the reduced width rule to the column headers.

Hopefully, that clears any confusion, any questions please let me know.

src/styles/semantic/themes/livingstone/globals/site.variables
```diff
@@ -428,3 +428,11 @@
 @searchBarFixedZIndex: 999;
 @searchBarFixedPadding: @20px 0;
 @searchBarFixedBoxShadow: 0 -(@4px) @9px -(@6px) rgba(0, 0, 0, 0.29);
+
+/*-------------------
+   Horizontal Gutter
+--------------------*/
+
+/* Minimum Gutter is used to determine  the maximum container width for a given device */
```
Jonnygwi: `/* Minimum Gutter is used to determine  the maximum container width for a given device */` ->
`/* Minimum Gutter is used to determine the maximum container width for a given device */`

src/styles/semantic/themes/livingstone/modules/modal.variables
```diff
@@ -53,3 +53,13 @@
 /* Size Widths */
 
 /* Derived Responsive Sizes */
+
+/* With the horizontal gutters */
+@reducedHeadingWidth:  ~"calc(100% - 30px)";
```
Jonnygwi: `@reducedHeadingWidth: ~"calc(100% - 30px)";`

src/styles/semantic/themes/livingstone/modules/modal.variables
```diff
@@ -53,3 +53,13 @@
...
+@withHorizontalCloseIconTop: @25px;
+@withHorizontalCloseIconRight: @25px;
+
+/*
+  480px = (ContainerWidth / 2) - Padding width
+*/
+@withHorizontalCloseIconLeft: ~"calc(50% + 480px)";
```
Jonnygwi: Turn this into a variable please `480px`

src/styles/semantic/themes/livingstone/globals/site.variables
stephenkr: I'm seeing double

src/styles/semantic/themes/livingstone/modules/modal.variables
stephenkr: Will do

src/styles/semantic/themes/livingstone/globals/site.variables
```diff
@@ -429,3 +429,13 @@
...
 @searchBarFixedBoxShadow: 0 -(@4px) @9px -(@6px) rgba(0, 0, 0, 0.29);
+
+/*-------------------
+   Horizontal Gutter
+--------------------*/
+
+/* Minimum Gutter is used to determine the maximum container width for a given device */
+
+@horizontalGutterMaxWidth: 1120px;
+// Must be defined in pixels for accuracy
```
Jonnygwi: This comment isn't necessary

src/styles/semantic/themes/livingstone/modules/modal.variables
```diff
@@ -53,3 +53,11 @@
 /* Size Widths */
 
 /* Derived Responsive Sizes */
+
+/* With the horizontal gutters */
+@reducedHeadingWidth: ~"calc(100% - 30px)";
```
Jonnygwi: Can we use `@30px` instead?

stephenkr: Yep i think that'll be fine.

src/styles/semantic/themes/livingstone/globals/site.variables
stephenkr: I think it might be, its important that this value is not replaced by `@40px`, this will change the positioning of the icon since `@40px` returns to 40px equivalent in rem.

Jonnygwi: Ahh okay okay

src/styles/semantic/themes/livingstone/modules/modal.overrides
```diff
@@ -20,6 +20,43 @@
...
 
+/* With Horizontal gutters */
+.ui.fullscreen.modal.with-horizontal-gutters {
+
+  .ui.container > .ui.header {
+    width: ~"calc(100% - 30px)";
+  }
+
+  &:not(.with-heading) {
+    @media (max-width: 599px) {
```
red-runner: Why `@media` instead of `@media screen and` ? 🤔

src/styles/semantic/themes/livingstone/modules/modal.overrides
```diff
@@ -20,6 +20,43 @@
...
+
+    @media (min-width: 600px) {
+
+      .grid-column-1 .ui.header {
+        width: ~"calc(100% - 30px)";
+      }
+    }
+  }
+
+  > i.icon {
```
red-runner: Is this `i`  icon tag necessary in this case?

src/styles/semantic/themes/livingstone/modules/modal.overrides
stephenkr: I'll add in the screen requirement

src/styles/semantic/themes/livingstone/modules/modal.overrides
stephenkr: Good to be consistent, its expected that the class icon will be assigned to the element `i`

src/components/media/Gallery/component.js
```diff
@@ -28,7 +28,10 @@ export const Component = ({ heading, images, trigger }) => (
       <Grid columns={2} stackable>
         <GridRow>
           {images.map(({ label, ...otherImageProps }, index) => (
-            <GridColumn key={buildKeyFromStrings(label, index)}>
+            <GridColumn
+              className={`grid-column-${index}`}
```
marcdavi-es: I don't think this is necessary? 

We can use CSS `:first-child` and `nth-child(1)`, no?

marcdavi-es: I don't think this is necessary? 

We can use CSS `:first-child` and `nth-child(1)`, no?

marcdavi-es: I don't think this is necessary? 

We can use CSS `:first-child` and `nth-child(1)`, no?

marcdavi-es: I don't think this is necessary? 

We can use CSS `:first-child` and `nth-child(1)`, no?

marcdavi-es: I don't think this is necessary? 

We can use CSS `:first-child` and `nth-child(1)`, no?

PR #390, opened 2018-10-11
TE-1183: only apply defaultValue if not null for the `Submenu`

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1183)

Turns out just setting `defaultValue` to null wasn't enough to remove the error, this update will fix that

### Before
screen shot 2018-10-15 at 10 37 12

### After
We get no error

screen shot 2018-10-15 at 10 38 33


src/components/elements/Submenu/component.js
```diff
@@ -38,6 +37,7 @@ export const Component = ({
     trigger={children}
     upward={willOpenAbove}
     {...getValueProperty(isSelectedDisabled)}
+    {...getDefaultValue(children, items, isSelectedDisabled)}
```
marcdavi-es: This is becoming increasingly confusing. 

Why are we spreading returned objects into props rather than using standard prop syntax?

stephenkr: I chose this approach to ensure the `defaultValue` prop would not be applied if the `value` prop existed.

I do see your point on this becoming confusing, i'll refactor to simply this.

PR #391, opened 2018-10-15
TE-1209 Refactor ShowOnMobile and ShowOnDesktop

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1209)

Fixes an issue where responsive wrapper wasn't correctly rendering after being rendered by the server.

### Any other notes
- Big PR but if you go through commit by commit it should all make sense. 
- A lot of snapshots which make up the bulk of lines added.
- Responsive wrappers rely on css instead of js and more importantly `window.innerWidth`

#### Header
kapture 2018-10-15 at 12 14 57
kapture 2018-10-15 at 12 14 01

#### Pictures
kapture 2018-10-15 at 12 16 59
kapture 2018-10-15 at 12 16 18

#### Rates
kapture 2018-10-15 at 12 19 20
kapture 2018-10-15 at 12 18 33

#### RoomType
image
image

#### HomepageHero
image
image

#### Summary
image
image


src/components/layout/ShowOnDesktop/component.js
```diff
@@ -1,25 +1,25 @@
...
-  <Responsive as={parent} {...parentProps} minWidth={600}>
-    {children}
-  </Responsive>
-);
+export const Component = ({ children, parent, parentProps }) =>
+  React.createElement(
+    parent,
+    {
+      ...parentProps,
+      className: getClassNames(parentProps.className, 'mobile-hidden'),
```
stephenkr: I think its cleaner to have a class that will only be applied if the desktop viewport is present, such as `desktop-shown`

If we have `ShowOnTablet`, we will need to update this code to hide on the tablet and any other responsive components we have.
Also adding these classes even if SemanticUi doesnt have support for them should be easy enough since the media query variables are exposed.

src/components/property-page-widgets/RoomType/component.js
```diff
@@ -108,9 +108,7 @@ const Component = ({
                 <Card.Description>
                   {getNightPriceMarkup(nightPrice)}
                 </Card.Description>
-                <ShowOnMobile>
-                  <Divider />
-                </ShowOnMobile>
+                <ShowOnMobile parent={Divider} />
```
stephenkr: Instead of using `parent`, can we rely on `as` to be consistent with SemanticUi's implementation?

src/components/media/Thumbnail/component.js
```diff
@@ -17,8 +17,9 @@ export const Component = ({
   isSquare,
   label,
   size,
+  className,
```
stephenkr: Would it be better to just have `ShowOnDesktop` wrap the element with a `div`?

And the div would have `<div className="desktop-shown">`

It would remove the need to have the extra `className` prop, although it would slightly change our component trees and may affect certain CSS selectors.

PR #392, opened 2018-10-17
TE-1216 Improve the SearchForm when displayed fixed

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1216)

Implements the new `PropertyPageSearchBar` component and its unit tests. It will serve as a responsive wrapper for the `SearchBar` component with the particularity of being displayed always as fixed. 

### Any other notes

* The wrapper of the `SearchBar` component was also slightly modified in order to fix unintended width behaviour. 

#### Before

screen shot 2018-10-17 at 12 29 18

### After

screen shot 2018-10-17 at 15 33 12

### Expected behaviour

propertypagesearchbar

### Coverage

screen shot 2018-10-17 at 15 51 56

### DIFF
--- a/src/components/general-widgets/SearchBar/component.js
+++ b/src/components/general-widgets/SearchBar/component.js
@@ -58,26 +58,28 @@ export class Component extends PureComponent {
             'is-fixed': isFixed,
           })}
         >
-          <HorizontalGutters as={Grid}>
-            <GridRow verticalAlign="middle">
-              <GridColumn width={5}>{summaryElement}</GridColumn>
-              <GridColumn floated="right" width={7}>
-                {isDisplayedAsModal ? (
-                  searchBarAsModal
-                ) : (
-                  <Form onSubmit={this.handleSubmit}>
-                    <Form.Group>
-                      {getFormFieldMarkup(
-                        this.props,
-                        this.persistInputChange,
-                        false,
-                        true
-                      )}
-                    </Form.Group>
-                  </Form>
-                )}
-              </GridColumn>
-            </GridRow>
+          <HorizontalGutters>
+            <Grid>
+              <GridRow verticalAlign="middle">
+                <GridColumn width={5}>{summaryElement}</GridColumn>
+                <GridColumn floated="right" width={7}>
+                  {isDisplayedAsModal ? (
+                    searchBarAsModal
+                  ) : (
+                    <Form onSubmit={this.handleSubmit}>
+                      <Form.Group>
+                        {getFormFieldMarkup(
+                          this.props,
+                          this.persistInputChange,
+                          false,
+                          true
+                        )}
+                      </Form.Group>
+                    </Form>
+                  )}
+                </GridColumn>
+              </GridRow>
+            </Grid>
           </HorizontalGutters>
         </div>
       );
@@ -124,15 +126,16 @@ Component.defaultProps = {
   summaryElement: null,
   willDropdownsOpenAbove: false,
 };
-/* eslint react/no-unused-prop-types: 0 */
 Component.propTypes = {
   /**
    * A function called for each day to be displayed in the DateRangePicker. Returning true blocks that day in the date range picker.
    * @param   {Moment}  day - The day to test.
-   * @returns{boolean}     - Is the day blocked.
+   * @returns {boolean}     - Is the day blocked.
    */
+  // eslint-disable-next-line react/no-unused-prop-types
   getIsDayBlocked: PropTypes.func,
   /** The options which the user can select in the guests field. */
+  // eslint-disable-next-line react/no-unused-prop-types
   guestsOptions: PropTypes.arrayOf(
     PropTypes.shape({
       /** The visible text for the option. */
@@ -150,8 +153,10 @@ Component.propTypes = {
   /** Is the Search Bar fixed to the bottom of the window */
   isFixed: PropTypes.bool,
   /** Is Search Bar showing the Property Summary info. */
+  // eslint-disable-next-line react/no-unused-prop-types
   isShowingSummary: PropTypes.bool,
   /** The options which the user can select in the location field. */
+  // eslint-disable-next-line react/no-unused-prop-types
   locationOptions: PropTypes.arrayOf(
     PropTypes.shape({
       /** The visible text for the option. */
@@ -178,6 +183,7 @@ Component.propTypes = {
    */
   onSubmit: PropTypes.func,
   /** The Search Button the Search Bar displays. */
+  // eslint-disable-next-line react/no-unused-prop-types
   searchButton: PropTypes.node,
   /** The element to display in the fixed container */
   summaryElement: PropTypes.node,
--- a/src/components/property-page-widgets/PropertyPageSearchBar/Readme.md
+++ b/src/components/property-page-widgets/PropertyPageSearchBar/Readme.md
@@ -1,0 +1,23 @@
+
+
+```jsx
+const { guestsOptions } = require('./mock-data/options');
+
+// Please set this to true to see the example
+const displayAsFixed = false;
+
+<PropertyPageSearchBar
+  isFixed={displayAsFixed}
+  guestsOptions={guestsOptions}
+  searchButton={
+  <Button isCompact isRounded>
+    Check Availability
+  </Button>
+  }
+  summaryElement={<div>Property Information</div>}
+  modalTrigger={<Button isPositionedRight isRounded isCompact>Check Availability</Button>}
+  modalSummaryElement={
+    <div>Property information for mobile modal</div>
+  }
+/> 
+```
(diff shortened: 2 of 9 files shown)


src/components/general-widgets/SearchBar/component.js
```diff
@@ -58,26 +58,28 @@ export class Component extends PureComponent {
...
-                        false,
-                        true
-                      )}
-                    </Form.Group>
-                  </Form>
-                )}
-              </GridColumn>
-            </GridRow>
+          <HorizontalGutters>
+            <Grid>
```
red-runner: This changed the width of the wrapper 
From: 
`width: calc(100% + 2rem)`
To:
`width: 100%`

The previous behaviour was causing the wrapper to bleed out of the screen

stephenkr:
@red-runner just wondering, how does this look with the location dropdown in the search bar?

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -442,3 +442,23 @@ blockquote.ui.quote .row {
...
+  }
+}
+
+/* Hide on mobile screens */
+@media only screen and (max-width: @computerBreakpoint) {
+
+  .property-page-search-desktop {
+    display: none !important;
+  }
+}
```
stephenkr: Always good to have an empty new line at the end of the document, removes diff noise

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -442,3 +442,23 @@ blockquote.ui.quote .row {
     display: none !important;
   }
 }
+
+/*----------------------
+    Property Page Search
+-----------------------*/
+
+/* Show on mobile screens */
+@media only screen and (max-width: @computerBreakpoint) {
```
stephenkr: Cleaner to nest these queries and selectors under a parent class like `.property-page-search`, just to keep this specific to the PropertyPageSearch component

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,98 @@
...
+  <Fragment>
+    <ShowOnDesktop
+      parent="div"
+      parentProps={{ className: 'property-page-search-desktop' }}
+    >
+      <SearchBar
+        guestsOptions={guestsOptions}
+        isFixed={isFixed}
+        searchButton={
+          <Button isCompact isRounded>
```
stephenkr: This should be exposed as a prop

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,98 @@
...
+            Check Availability
+          </Button>
+        }
+        summaryElement={<div>Property information</div>}
```
stephenkr: This also should be exposed as a prop

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,98 @@
...
+        PropTypes.string,
+      ]),
+    })
+  ).isRequired,
+  /**
+   * Is the Search Bar fixed to the bottom of the window
+   * Used for demo purposes
+   * @ignore
+   */
+  /** Is the Search Bar fixed to the bottom of the window */
```
stephenkr: This comment is not needed

src/components/property-page-widgets/PropertyPageSearchBar/component.spec.js
```diff
@@ -0,0 +1,25 @@
...
+
+const getPropertyPageSearchBar = props =>
+  mount(<PropertyPageSearchBar guestsOptions={guestsOptions} {...props} />);
+
+describe('<PropertyPageSearchBar />', () => {
+  it('should have the right structure', () => {
+    const actual = getPropertyPageSearchBar();
+
+    expect(actual).toMatchSnapshot();
+  });
```
stephenkr: You should also test the props and ensure the props are correctly being applied

src/styles/semantic/themes/livingstone/globals/site.overrides
Jonnygwi: linting tests should fail in this case. @red-runner try running `npm run pre-commit` and see if it passes

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,98 @@
...
+    <ShowOnDesktop
+      parent="div"
+      parentProps={{ className: 'property-page-search-desktop' }}
+    >
+      <SearchBar
+        guestsOptions={guestsOptions}
+        isFixed={isFixed}
+        searchButton={
+          <Button isCompact isRounded>
+            Check Availability
```
Jonnygwi: Don't forget this text needs to be brought out into a constants and turned into an optional prop so it can be translated

src/components/property-page-widgets/PropertyPageSearchBar/component.js
Jonnygwi: Same here: https://github.com/lodgify/lodgify-ui/pull/392/files#r226242961

src/styles/semantic/themes/livingstone/globals/site.overrides
red-runner: @Jonnygwi I ran `npm run test` instead of `npm run pre-commit`; it was indeed failing 💥

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,91 @@
...
+    })
+  ).isRequired,
+  /**
+   * Is the Search Bar fixed to the bottom of the window
+   * Used for demo purposes
+   * @ignore
+   */
+  isFixed: PropTypes.bool,
+  /** Is Search Bar showing the Property Summary info. */
+  isShowingSummary: PropTypes.bool,
```
Jonnygwi: I'm confused why this is here if it's not being used

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,91 @@
...
+  isFixed: true,
+  isShowingSummary: false,
+  searchButton: (
+    <Button isPositionedRight isRounded>
+      {CHECK_OUR_AVAILABILITY}
+    </Button>
+  ),
+  summaryElement: null,
+};
+/* eslint react/no-unused-prop-types: 0 */
```
Jonnygwi: We should only use `// eslint-disable-next-line react/no-unused-prop-types` so that we know which prop type is being ignored

src/components/property-page-widgets/PropertyPageSearchBar/component.js
red-runner: @Jonnygwi I think I got confused with this prop. I removed it from everywhere.

src/components/property-page-widgets/PropertyPageSearchBar/component.js
red-runner: @Jonnygwi I modified the `SearchBar` component with your suggestion where was needed. I removed the other one accordingly.

src/components/property-page-widgets/PropertyPageSearchBar/Readme.md
```diff
@@ -0,0 +1,23 @@
...
+  <Button isCompact isRounded>
+    Check Availability
+  </Button>
+  }
+  summaryElement={<div>Property Information</div>}
+  modalTrigger={<Button isPositionedRight isRounded isCompact>Check Availability</Button>}
+  modalSummaryElement={
+    <div>Property information for mobile modal</div>
+  }
+/>
```
marcdavi-es: All the 'markup' type props in this example are inconsistent. Some are multilined, some are inlined and the indentation is bad.

Please make consistent.

src/components/property-page-widgets/PropertyPageSearchBar/component.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const Component = ({
+  guestsOptions,
+  isFixed,
+  summaryElement,
+  modalTrigger,
+  modalSummaryElement,
+  searchButton,
+}) => (
+  <div className="property-page-searchbar">
```
marcdavi-es: Shouldn't need this wrapping `div` - wrap this in a `Fragment`

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -442,3 +442,20 @@ blockquote.ui.quote .row {
...
+
+    .show-on-mobile {
+      display: block !important;
+    }
+
+    .show-on-desktop {
+      display: none !important;
+    }
+  }
+}
```
marcdavi-es: I don't think we need any of these styles? Just use `ShowOnMobile` and `ShowOnDesktop`

PR #395, opened 2018-10-19
TE-1233: Update Location prop requirements

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1233)

Ensure `locationDescription` and `transportOptions` are optional

### Before
screen shot 2018-10-19 at 16 26 03

### After
screen shot 2018-10-19 at 16 26 27
screen shot 2018-10-19 at 16 26 22
screen shot 2018-10-19 at 16 26 14


src/components/property-page-widgets/Location/component.spec.js
```diff
@@ -215,6 +217,42 @@ describe('<Location />', () => {
     });
   });
 
+  describe('the `locationDescription`', () => {
```
Jonnygwi: As we are slowing adding snapshots across our components we should probably do the same with this one instead. We can save on a lot of lines! :D

stephenkr: Adding them now.

src/components/property-page-widgets/Location/component.spec.js
```diff
@@ -1,25 +1,13 @@
 import React from 'react';
 import { shallow } from 'enzyme';
```
marcdavi-es: Let's use `mount` for a safer snapshot.

PR #399, opened 2018-10-23
TE-1283 Add theming to menu item underlines

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1283)

The menu items underline is determined by the theme's primary colour.

### Any other notes
kapture 2018-10-22 at 16 07 24

kapture 2018-10-23 at 11 47 27

### DIFF
--- a/src/styles/semantic/themes/livingstone/collections/menu.overrides
+++ b/src/styles/semantic/themes/livingstone/collections/menu.overrides
@@ -123,6 +123,7 @@
 
         &.active {
           border-bottom: @activeItemUnderlineBorder;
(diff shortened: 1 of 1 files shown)


src/styles/semantic/themes/livingstone/collections/menu.overrides
```diff
@@ -122,7 +122,7 @@
         }
 
         &.active {
-          border-bottom: @activeItemUnderlineBorder;
+          border-bottom: var(@activeItemUnderlineBorder, @themeActionColorDefault);
```
marcdavi-es: I don't think this is right.

Should match lines 293 and 294, right?

PR #403, opened 2018-10-23
TE-1263: Refactor FullBleed & Promotion to use ResponsiveImage

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1263)

Update the `FullBleed` and the `Promotion` components to use the `ResponsiveImage` instead of using the background image with CSS

### Other notes
- There is no visible difference between the after version and the before version
- This update relies on `object-fit`, this CSS rule is not supported by Internet Explorer, so I have added a fallback to handle browsers that don't support this rule. More information on this below.
- The promotion specs have been refactored to use snapshot testing

### Before - full bleed in PropertyPageHero
screen shot 2018-10-23 at 16 41 50

### After - full bleed in PropertyPageHero
screen shot 2018-10-23 at 16 41 24

__Without Object-fit being supported or applied__
screen shot 2018-10-23 at 17 16 38

### Before - Promotion
screen shot 2018-10-23 at 17 18 45

__As stacked__
screen shot 2018-10-23 at 17 20 10

### After - Promotion
screen shot 2018-10-23 at 17 19 50

__As stacked__
screen shot 2018-10-23 at 17 20 29

### The Object-fit fallback
The solution is simply:
- Set the image `min-width` and `min-height` to 100%
- Use `position absolute` with `transform` to center the image in the container
- Set the `width` to 100% to fit the parent container
- Use a feature query to disable the fallback and apply the `object-fit`, also good to note that feature queries are not supported by Internet Explorer

#### Here is a preview of how it looks:

__With the fallback__
- The blue highlight represents the image element
screen shot 2018-10-24 at 15 45 06

__With object-fit: cover__
screen shot 2018-10-24 at 15 45 22

#### Preview with a portrait:

__With the fallback__
- The blue highlight represents the image element
screen shot 2018-10-24 at 15 47 56

__With object-fit: cover__
screen shot 2018-10-24 at 15 47 34


src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -459,3 +459,20 @@ blockquote.ui.quote .row {
...
+    Responsive Image
+-----------------------*/
+.responsive-image {
+
+  &.is-container-filled {
+    height: 100%;
+    width: 100%;
+
+    .ui.fluid.image {
+      object-fit: cover;
```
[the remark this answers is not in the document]
stephenkr: Hi @Jonnygwi 

Yep that's a concern with this rule, we could look into a fallback for IE.

Infact we could rely on using `@supports`, this is supported on the same browsers that support the `object-fit` rule.

So a possible fix could be something like:
```
@supports (object-fit: cover) {
    .with-rule {
         object-fit: cover;

         // disable any IE specific fixes
    }
}

```

But heres a screenshot of how the `PropertyPageHero` will look without the `object-fit` rule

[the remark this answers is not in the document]
Jonnygwi: We are currently doing graceful loading which is the step before lazy loading so there isn't a need for https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API right now.

I had this conversation with @marcdavi-es and we agreed that it would be best to have all the graceful loading contained to one component.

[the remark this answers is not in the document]
stephenkr: Ah ok.

What do you think about the IE preview? 

Actually, a possible fix could be the inherit the background image using a data prop, this way we can utilize CSS background size.

This is widely supported.

Such as:
```JSX:
<ResponsiveImage as={LazyLoad} data-image="https://">

</ResponsiveImage>
```

```LESS:
.lazy-load.responsive-image {

  &:after {
    background-image: attr(data-image);
    background-size: cover;
  }
}
```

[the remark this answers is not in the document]
stephenkr: Ok discussing with Julio, I have proposed a possible solution that will work for IE that Julio is ok with.

Simply the image will have a min height and width of 100% of the container, and using `position: absolute` and `transform` the image will be centered to the container.

Then using `@supports`, we will determine if the `object-fit` is to be applied or the fallback solution.

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -474,3 +474,38 @@ blockquote.ui.quote .row {
...
+
+    /* Object fit cover Fallback */
+    .ui.fluid.image {
+      min-width: 100%;
+      width: 100%;
+      min-height: 100%;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
```
Jonnygwi:
```
      top: 50%;
      left: 50%;
      transform: translate(-50%, -50%);
```
These three should be turned into variables

src/components/media/ResponsiveImage/component.js
```diff
@@ -72,6 +80,8 @@ Component.propTypes = {
   imageUrl: PropTypes.string,
   /** Whether to render the image as an avatar */
   isAvatar: PropTypes.bool,
+  /** Whether to fill the image container */
+  isFilledContainer: PropTypes.bool,
```
marcdavi-es: Do we need a new prop for this? Can't we extend the behaviour of `isFluid`?

stephenkr: Yep true, i'll update this.

stephenkr:
Hi @Jonnygwi 

There were some changes applied, can you give this another review?

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -474,3 +474,38 @@ blockquote.ui.quote .row {
...
+      .ui.fluid.image {
+        position: static;
+        object-fit: cover;
+        height: inherit;
+        width: inherit;
+        transform: none;
+      }
+    }
+  }
+}
```
marcdavi-es: These styles should probably live in `image.overrides`?

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -474,3 +474,38 @@ blockquote.ui.quote .row {
     }
   }
 }
+
+/*----------------------
+    Responsive Image
+-----------------------*/
+.responsive-image {
```
marcdavi-es: More specificity? `picture.responsive-image`

src/styles/semantic/themes/livingstone/globals/site.overrides
stephenkr: Good point, moving them now.

src/styles/semantic/themes/livingstone/globals/site.overrides
stephenkr: Ok

PR #408, opened 2018-10-24
TE-1276: Import Semantic components by path

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1276)

Import each Semantic UI react components by path. 

### other notes
The Semantic components are imported from the `/commonjs` folder because Jest does not yet have support for ES modules.
- The support for this is being discussed here `https://github.com/facebook/jest/issues/4842`
- The component is visually and functionally the same no matter if imported with ES modules or CommonJS

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "size-limit": [
     {
-      "limit": "321 KB",
+      "limit": "319 KB",
       "path": "lib/index.js"
     }
   ],
--- a/src/components/collections/Footer/component.js
+++ b/src/components/collections/Footer/component.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Menu } from 'semantic-ui-react';
+import Menu from 'semantic-ui-react/dist/commonjs/collections/Menu';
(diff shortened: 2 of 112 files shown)


stephenkr:
Hi @marcdavi-es 

So running `npm run build` and then `npm run size`, the results are:
```
Package size: 316.54 KB
Size limit:   321 KB
With all dependencies, minified and gzipped
```

On point 1 on using the path imports in the specs, is it not better to be consistent with the non-test files?

I'll update the size limit in the package.json.

stephenkr:
Hi @marcdavi-es 

Sorry for not including the before size, the before and after sizes are shown below.

Actually looking at the information below, it appears that using path imports is heavier than the current size in production. This must be because the `commonjs` version of the Semantic UI components is bigger in size.

Although when running this test before with production, this change had a smaller build size.

**Should we close this pull request since there is no advantage?**

### Before
```
Package size: 251.82 KB
Size limit:   256 KB
```

### After
```
Package size: 254.13 KB
Size limit:   256 KB
```

stephenkr:
Closing in favor of using a Babel plugin to transform the Semantic UI react imports.

PR #414, opened 2018-10-29
TE-1303 Thumbnail to use ResponsiveImage

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1303)

Gives the Thumbnail the capability to gracefully load

#### Thumbnail
**Before**
kapture 2018-10-29 at 12 55 45
**After**
kapture 2018-10-29 at 12 54 54

#### Pictures
image

#### ResponsiveImage
image

image

### DIFF
--- a/src/components/media/ResponsiveImage/component.js
+++ b/src/components/media/ResponsiveImage/component.js
@@ -31,6 +31,8 @@ export class Component extends PureComponent {
       placeholderImageUrl,
       imageUrl,
       isFluid,
+      hasRoundedCorners,
+      isCircular,
     } = this.props;
 
     const imageProps = {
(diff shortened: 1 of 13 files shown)


src/components/media/ResponsiveImage/component.js
```diff
@@ -43,6 +45,8 @@ export class Component extends PureComponent {
       <picture
         className={getClassNames('responsive-image', {
           'is-fluid': isFluid,
+          rounded: hasRoundedCorners,
+          circular: isCircular,
```
stephenkr: For the two new classes being added, should we keep with the structure of prefixing `is-`, such as:
```
'is-rounded': hasRoundedCorners,
'is-circular': isCircular
```

PR #422, opened 2018-11-02
TE-1366: Add form success and error message props

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1366)

Exposes error and success message props that if passed into the `Form` will display a message

### Other notes
- I will apply the `successMessage` and `errorMessage` props to the other forms once this PR has been approved and merged

### Preview of the success message
screen shot 2018-11-02 at 16 26 57

### Preview of the error message
screen shot 2018-11-02 at 16 27 02

### Impact on the bundle size
- Minor increase due to the message styles now being imported from Semantic UI

#### Before
```
Package size: 255.1 KB
Size limit:   256 KB
```

#### After
```
Package size: 255.97 KB
Size limit:   256 KB
```


src/components/collections/Form/Readme.md
```diff
@@ -198,3 +198,29 @@ const dropdownOptions = [{
...
+```jsx
+<Form
+  headingText="Welcome"
+  submitButtonText="Sign up"
+  errorMessage="Request Failed. Please try again."
+>
+  <TextInput label="Name" name="name" />
+  <TextInput label="Password" name="password" type="password"/>
+</Form>
+```
```
marcdavi-es: I think we should have a `States` section in the example file with `Success` and `Error` subsections, like e.g. https://lodgify.github.io/lodgify-ui/#/Inputs/TextInput

stephenkr: Ok, I'll add in the section.

src/components/collections/Form/component.js
```diff
@@ -90,10 +103,12 @@ export class Component extends PureComponent {
 Component.displayName = 'Form';
 
 Component.defaultProps = {
+  errorMessage: '',
```
Jonnygwi: Are we not including default string for these text props?
Are we not following the normal `<something>Text` prop naming? like how we've done `headingText` for example

src/components/collections/Form/component.js
```diff
@@ -115,6 +132,8 @@ Component.propTypes = {
   onSubmit: PropTypes.func,
   /** The text to display on the submit button. */
   submitButtonText: PropTypes.string,
+  /** The message to display when the form is successful */
+  successMessage: PropTypes.string,
```
Jonnygwi: Same for here https://github.com/lodgify/lodgify-ui/pull/422/files#r230751292

PR #429, opened 2018-11-12
TE-1389: The Promotion appears as expected if nested under Grid

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1389)

Fixes the rendering of the Promotion component for the mobile viewport when the component is nested in a `Grid`

### Any other notes
- Also, when the Promotion is shown on mobile, the discount section will now have the padding correctly applied

### Why this was occurring
- Conflicting CSS rules were being applied to a nested Grid in the Promotion component when the component was being rendered in TemplatesSSR, this update will override these conflicting CSS rules.

In the following previews, the promotion component is nested in a `Grid`

### Before
screen shot 2018-11-12 at 17 32 09

### After
screen shot 2018-11-12 at 17 34 21

### DIFF
--- a/src/styles/semantic/themes/livingstone/elements/segment.overrides
+++ b/src/styles/semantic/themes/livingstone/elements/segment.overrides
@@ -131,6 +131,11 @@
   @media @tabletScreen {
     padding: 0;
 
+    .ui.stackable.grid {
+      margin-left: 0 !important;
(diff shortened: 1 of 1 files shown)


stephenkr:
Hi @marcdavi-es 

I have updated the description and the commit message to be more specific on what is being changed.

PR #430, opened 2018-11-12
TE-1410 Implement `RatingInput` component

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1410)

Implements the `RatingInput` component which allows the user to input a value based on stars.
### Any other notes

* The implementation of this component required a refactor of the `InputController` component as well as its `getValue` util. 

* The reason of the change being the implementation of the `adaptOnChangeEvent` prop, which gives to every children of `InputController` the power to parse `event` values according to their needs. 

ratinginput


src/components/inputs/RatingInput/component.js
```diff
@@ -0,0 +1,27 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { Rating } from 'semantic-ui-react';
+
+/**
+ * A rating input that allows the user to input a value based on stars.
+ */
+// eslint-disable-next-line jsdoc/require-jsdoc
+export const Component = ({ iconSize, maxRating }) => (
+  <Fragment>
```
marcdavi-es: What does this `Fragment` do?

src/components/inputs/RatingInput/component.js
```diff
@@ -0,0 +1,27 @@
...
+    <Rating maxRating={maxRating} size={iconSize} />
+  </Fragment>
+);
+
+Component.displayName = 'RatingInput';
+
+Component.defaultProps = {
+  iconSize: 'small',
+  maxRating: 5,
```
marcdavi-es: Didn't we agree that we are not exposing `maxRating` ?

red-runner: @marcdavi-es Yes, I got confused, apologies

src/components/inputs/RatingInput/component.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from 'prop-types';
+import { Rating } from 'semantic-ui-react';
+
+import { InputController } from '../InputController';
+
+/**
+ * A rating input that allows the user to input a value based on stars.
+ */
+// eslint-disable-next-line jsdoc/require-jsdoc
+export const Component = ({ iconSize, error, name, onChange }) => (
```
stephenkr: Can you sort these props alphabetically?

src/components/inputs/RatingInput/component.js
```diff
@@ -0,0 +1,44 @@
...
+  <InputController
+    error={error}
+    isValid={false}
```
stephenkr: Is this meant to be hardcoded to `false`?

src/components/inputs/RatingInput/component.js
```diff
@@ -0,0 +1,44 @@
...
+    name={name}
+    onChange={onChange}
+  >
+    <Rating maxRating={5} size={iconSize} />
```
stephenkr: Can we extract this out as a constant?

src/components/inputs/RatingInput/component.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A rating input that allows the user to input a value based on stars.
+ */
+// eslint-disable-next-line jsdoc/require-jsdoc
+export const Component = ({ iconSize, error, name, onChange }) => (
+  <InputController
+    error={error}
+    isValid={false}
+    name={name}
+    onChange={onChange}
```
stephenkr: I don't see how this works. Will the `Rating` component trigger the `onChange` method when a star is selected?

The `inputController` will only pass along the `onChange` method and it assumes the wrapper Component/Element will handle the `onChange`

src/components/inputs/RatingInput/component.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    const props = {
+      error: 'Shomething is odd',
+      iconSize: 'tiny',
+      name: 'RatingInput',
+    };
+    const wrapper = getRatingInput({ props });
+
+    expect(wrapper).toMatchSnapshot();
+  });
+
```
stephenkr: Can you add a spec to ensure `onChange` is getting triggered?

src/components/inputs/RatingInput/component.js
red-runner: @stephenkr Yes, this component doesn't need validation but `isValid` is required in InputController

src/components/inputs/RatingInput/component.js
red-runner: @stephenkr I followed the same pattern as the `Rating` component and we don't have a const there. Do you think it's necessary?

src/utils/get-argument-through/getArgumentThrough.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * @param {*} value
+ * @return {*}
```
marcdavi-es: I think we use `{any}` rather than `{*}` ?

src/utils/get-argument-through/getArgumentThrough.js
```diff
@@ -0,0 +1,5 @@
...
+ */
+export const getArgumentThrough = value => value;
```
marcdavi-es: I find this name difficult to understand. Maybe `returnFirstArgument`?

red-runner:
@marcdavi-es `getArgumentThrough ` comments are solved

src/utils/return-first-argument/returnFirstArgument.spec.js
```diff
@@ -0,0 +1,10 @@
+import { returnFirstArgument } from './returnFirstArgument';
+
+describe('returnFirstArgument', () => {
+  it('should return the value provided by the function argument', () => {
```
stephenkr: I'd rewrite this as `it('should only return the first argument', () => {`

src/utils/return-first-argument/returnFirstArgument.spec.js
```diff
@@ -0,0 +1,10 @@
...
+    const value = 'Last night a DJ saved my life';
+    const actual = returnFirstArgument(value);
+
+    expect(actual).toBe(value);
+  });
```
stephenkr: Could you add another spec to cover the case for when no arguments are passed into the function?

src/components/inputs/InputController/utils/getValue.js
```diff
@@ -1,11 +1,7 @@
...
- * @param  {Object|any} eventOrValue
+ * @param  {Object} event
  * @return {any}
  */
-export const getValue = eventOrValue => {
-  const nestedValue = get(eventOrValue, ['target', 'value']);
-
-  return !isUndefined(nestedValue) ? nestedValue : eventOrValue;
-};
+export const getValue = event => get(event, ['target', 'value']);
```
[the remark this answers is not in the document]
red-runner: @stephenkr `getValue` is the name it was used before, changing it would mean to switch 21 results across 7 files. Could we leave it for another separe PR?

red-runner:
Stephen's comments are solved

PR #431, opened 2018-11-13
TE-1406: Just load the image if it exists in the cache

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1406)

If the image is in the cache, then just show it and do not show the blurred image. 

### Any other notes
- Also removed the `isFluid` JsDoc param from the image markup utilities

### Preview of Gallery
kapture 2018-11-13 at 15 37 12

### Preview of the Homepage Hero
kapture 2018-11-13 at 15 35 00

### DIFF
--- a/src/components/media/ResponsiveImage/component.js
+++ b/src/components/media/ResponsiveImage/component.js
@@ -19,11 +19,27 @@ import { getPlaceholderImageMarkup } from './utils/getPlaceholderImageMarkup';
 // eslint-disable-next-line jsdoc/require-jsdoc
 export class Component extends PureComponent {
   state = {
-    shouldImageLoad: false,
     isImageLoaded: false,
+    isImageLoadedFromCache: false,
+    shouldImageLoad: false,
   };
 
-  componentDidMount = () => this.setState({ shouldImageLoad: true });
+  componentDidMount = () => {
+    const { imageUrl, placeholderImageUrl } = this.props;
+
+    if (!!placeholderImageUrl && !!imageUrl) {
+      const fullsizeImage = new Image();
+
+      fullsizeImage.src = imageUrl;
+
+      this.setState({
+        shouldImageLoad: true,
+        isImageLoadedFromCache: fullsizeImage.complete,
+      });
+    } else {
+      this.setState({ shouldImageLoad: true });
+    }
+  };
 
   handleImageLoad = () => this.setState({ isImageLoaded: true });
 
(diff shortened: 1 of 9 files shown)


src/components/media/ResponsiveImage/component.js
```diff
@@ -19,11 +19,27 @@ import { getPlaceholderImageMarkup } from './utils/getPlaceholderImageMarkup';
...
+    isImageLoadedFromCache: false,
+    shouldImageLoad: false,
   };
 
-  componentDidMount = () => this.setState({ shouldImageLoad: true });
+  componentDidMount = () => {
+    const { imageUrl, placeholderImageUrl } = this.props;
+
+    if (!!placeholderImageUrl && !!imageUrl) {
+      const fullsizeImage = new Image();
```
Jonnygwi: Will this still work if another component has loaded the image? As in will it use the already loaded one?

stephenkr:
Hi @marcdavi-es 

For the component to detect if the image is loaded from the cache, we need access to the client-side cache.

What you are seeing in the preview is the server side render rendering first a hidden image and once the client-side JavaScript is ready, the image is shown or the blurred placeholder depending on if the image exists in the cache already.

Originally I had the blurred image to be shown by default, but once the client-size render ran, the user would see a flash of the blurred image and then the actual image. It made sense to hide the image first and then show the image once the client side JavaScript was ready.

I've already shown Julio this behavior and he is ok with it.

src/components/media/ResponsiveImage/utils/getPlaceholderImageMarkup.js
```diff
@@ -31,13 +32,16 @@ export const getPlaceholderImageMarkup = ({
...
   placeholderImageUrl,
   shouldImageLoad,
   /* eslint-enable react/prop-types */
 }) => (
   <div
     className={getClassNames('image-with-placeholder-container', {
-      'has-blurred-children': !isImageLoaded,
+      'has-blurred-children': !isImageLoaded && !isImageLoadedFromCache,
+      'show-placeholder-container': shouldImageLoad,
+      'with-cached-image': isImageLoadedFromCache,
```
marcdavi-es: Rather than use classes and css, can't we use `isImageLoadedFromCache` in the check here https://github.com/lodgify/lodgify-ui/pull/431/files#diff-5f20f569d86a2305795e5368b21d7debR81 ?

If that check is unwieldy, let's isolate it in a little util.

stephenkr: So the class was applied here to help with transitioning in the image in a graceful way if the image was already loaded.

### Heres a preview with the transition of the cached image: 

kapture 2018-11-15 at 11 44 46

### And here is a preview without the transition:

kapture 2018-11-15 at 11 47 27

If you are ok with the transition approach, I'll improve this implementation to use a utility.

PR #434, opened 2018-11-14
TE-1423 block dates that are in the past

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1423)

Greys out the dates that are in the past.

#### Before
image

#### After
image

### DIFF
--- a/src/components/property-page-widgets/Availability/utils/isDayBlockedOrBeforeCurrentDate.js
+++ b/src/components/property-page-widgets/Availability/utils/isDayBlockedOrBeforeCurrentDate.js
@@ -1,0 +1,11 @@
+import moment from 'moment';
+import { isInclusivelyAfterDay } from 'react-dates';
+
+/**
+ * @typedef Moment
+ * @param   {Moment}   day
+ * @param   {Function} getIsDayBlocked
+ * @return  {boolean}
+ */
+export const isDayBlockedOrBeforeCurrentDate = (day, getIsDayBlocked) =>
+  !!day && (getIsDayBlocked(day) || !isInclusivelyAfterDay(day, moment()));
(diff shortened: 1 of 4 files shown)


src/components/property-page-widgets/Availability/utils/isDayBlockedOrBeforeCurrentDate.js
```diff
@@ -0,0 +1,11 @@
+import moment from 'moment';
+import { isInclusivelyAfterDay } from 'react-dates';
+
+/**
+ * @typedef Moment
+ * @param  {Moment} day
+ * @param  {Function} getIsDayBlocked
```
stephenkr: Can you better align these props? Such as:
```
@param {Moment}    day
@param {Function}   getIsDayBlocked
```

src/components/property-page-widgets/Availability/utils/isDayBlockedOrBeforeCurrentDate.js
```diff
@@ -0,0 +1,11 @@
...
+ * @return {boolean}
+ */
+export const isDayBlockedOrBeforeCurrentDate = (day, getIsDayBlocked) =>
+  day && (getIsDayBlocked(day) || !isInclusivelyAfterDay(day, moment()));
```
stephenkr: It would be better to convert `day` at the start of this condition into a boolean, such as `!!day && (` 

Or update the JsDoc return to indicate that `undefined` can also be returned.

src/components/property-page-widgets/Availability/utils/isDayBlockedOrBeforeCurrentDate.spec.js
```diff
@@ -0,0 +1,42 @@
...
+      const actual = isDayBlockedOrBeforeCurrentDate(day, getIsDayBlocked);
+
+      expect(actual).toMatchSnapshot();
+    });
+  });
+
+  describe('if day is not blocked', () => {
+    const getIsDayBlocked = () => false;
+
+    it('should return the right value when date is in the passed', () => {
```
stephenkr: `passed` to `past`

src/components/property-page-widgets/Availability/utils/isDayBlockedOrBeforeCurrentDate.js
Jonnygwi: I'll change this but I think we should try and find some linting for this. It's pretty inconsistent across the project and going back to make these lintable changes are a time sink.

PR #443, opened 2018-11-22
TE-1486: Set the PropertyPageSearchBar height to 85px

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1486)

Ensures the PropertyPageSearchBar has a height set to `85px` and tidy up the inner components to ensure no content wrapping occurs.

### Any other notes
- The Property page search bar container will be set to 85px, this is to provide whitespace after the footer on the property page.
- Tweaked the summary to truncate the location name and property name to remain within the parent container
- The Rating will be hidden at a certain breakpoint in the property page search bar to provide enough space for the localised text in the Summary to expand

#### Desktop view
screen shot 2018-11-22 at 12 52 19

#### Tablet view
screen shot 2018-11-22 at 12 52 34

#### Mobile view
screen shot 2018-11-22 at 14 39 37


src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -463,9 +473,25 @@ blockquote.ui.quote .row {
 -----------------------*/
 
 .property-page-searchbar {
+  height: @propertyPageSearchBarHeight;
 
   .ui.segments > .ui.segments {
     margin: 0;
+
+    @media only screen and (min-width: @smallestLargeComputerBreakpoint) and (max-width: (@smallestLargeComputerBreakpoint + 100)) {
```
Jonnygwi: Where did this `100` come from? If it comes from something then you should probably pull it out into a variable so it makes sense as to what this calculation is doing.

marcdavi-es: We also have a pattern for isolating media queries and managing them in one place... 

##### site.variables

```
/*-------------------
     Breakpoints
--------------------*/

@tinyBreakpoint: 400px;
@tabletBreakpoint: 600px;
@computerBreakpoint: 1000px;
@smallestComputerBreakpoint: (@tabletBreakpoint + 1px);

@tinyScreen: ~"only screen and (max-width: @{largestTinyScreen})";
@mobileScreen: ~"only screen and (max-width: @{largestMobileScreen})";
@tabletScreen: ~"only screen and (max-width: @{tabletBreakpoint})";
@computerScreen: ~"only screen and (min-width: @{smallestComputerBreakpoint})";
```

stephenkr: Yes good idea, I'll extract it out to a variable.

So the `100px` is an offset from the smallest large computer breakpoint. It is used to hide to the Review in the Summary when the search form is too close to the Summary.

src/styles/semantic/themes/livingstone/globals/site.variables
```diff
@@ -171,6 +171,7 @@
 @tabletBreakpoint: 600px;
 @computerBreakpoint: 1000px;
 @smallestComputerBreakpoint: (@tabletBreakpoint + 1px);
+@smallestLargeComputerBreakpoint: (@computerBreakpoint + 1px);
```
marcdavi-es: `smallestLarge` starts to sound a bit daft

Maybe we can use `widescreen` as a name for the next level up after `computer`?

src/styles/semantic/themes/livingstone/globals/site.overrides
stephenkr: I'll move the offset breakpoint to this section in `site.variables`

PR #446, opened 2018-11-22
TE-1468 RoomType image size

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1468)

Restricts the RoomType slideshow image sizes and adds a graceful fallback for IE 11. Hides the bullet navigation when there is only one image in the slideshow.

### Before
image

### After
** Modal **
screenshot 2018-11-22 at 12 46 10

** Standard **
screenshot 2018-11-22 at 12 46 15

** Mobile ** 
screenshot 2018-11-22 at 16 55 40

#### When only one image
screenshot 2018-11-22 at 12 46 31

#### IE 11
** Modal **
screenshot 2018-11-22 at 12 21 03

** Standard **
screenshot 2018-11-22 at 12 20 08

** Mobile ** 
screenshot 2018-11-22 at 16 46 53

### DIFF
--- a/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
+++ b/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Modal, List } from 'semantic-ui-react';
+import { size } from 'lodash';
 
 import { buildKeyFromStrings } from 'utils/build-key-from-strings';
 import { Amenities } from 'property-page-widgets/Amenities';
@@ -34,11 +35,15 @@ export const getModalContentMarkup = (
   slideShowImages,
   isUserOnMobile
 ) => (
-  <Modal.Content>
+  <Modal.Content className="room-type">
     <Heading>{name}</Heading>
     {!!ratingNumber && <Rating ratingNumber={ratingNumber} />}
     <Divider size="small" />
-    <Slideshow additionalClass="no-shadow" images={slideShowImages} />
+    <Slideshow
+      additionalClass="no-shadow"
+      images={slideShowImages}
+      isShowingBulletNavigation={size(slideShowImages) > 1}
+    />
     {!!description ? (
       <Paragraph>{description}</Paragraph>
     ) : (
--- a/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
+++ b/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
@@ -40,7 +40,6 @@ export const getModalContentMarkup = (
     {!!ratingNumber && <Rating ratingNumber={ratingNumber} />}
     <Divider size="small" />
     <Slideshow
-      additionalClass="no-shadow"
       images={slideShowImages}
       isShowingBulletNavigation={size(slideShowImages) > 1}
     />
--- a/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
+++ b/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
@@ -35,7 +35,7 @@ export const getModalContentMarkup = (
   slideShowImages,
   isUserOnMobile
 ) => (
-  <Modal.Content className="room-type">
+  <Modal.Content className="has-room-type-gallery">
     <Heading>{name}</Heading>
     {!!ratingNumber && <Rating ratingNumber={ratingNumber} />}
     <Divider size="small" />
--- a/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.spec.js
+++ b/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.spec.js
@@ -52,7 +52,7 @@ describe('getModalContentMarkup', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
-  describe('`props.ratingNumber` if null', () => {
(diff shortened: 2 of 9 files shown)


src/styles/semantic/themes/livingstone/views/card.overrides
```diff
@@ -113,3 +113,37 @@
...
+
+/* Room Type Image Gallery */
+.has-room-type-gallery {
+
+  .image-gallery {
+
+    .image-gallery-content {
+
+      .image-gallery-image {
+        height: @roomTypeSlideshowImageHeight;
```
stephenkr: Is it ok to set a fixed height for the image gallery in the RoomType? What if the title is long enough to expand the size of the card?

src/styles/semantic/themes/livingstone/views/card.overrides
```diff
@@ -113,3 +113,37 @@
...
+
+        > img {
```
stephenkr: Could we deduplicate this fallback to a class defined in `site.overrides`? We could then extend the class to where the fallback is required.

src/styles/semantic/themes/livingstone/views/card.variables
```diff
@@ -94,3 +94,5 @@
 /* Sizes */
 
 /* Colored */
+
+@roomTypeSlideshowImageHeight: 180px;
```
stephenkr: Can you add a comment for the RoomType variables?

Jonnygwi: I don't know what this means. A comment on what?

stephenkr: For the section, we follow a comment structure. Such as:
```
/* Room Type image gallery */
@roomTypeSlideshowImageHeight: 180px;
```

src/styles/semantic/themes/livingstone/views/card.overrides
Jonnygwi: I tried using the ruleset variables less provides but our linting and less versions aren't matched up so that is currently blocked on https://youtrack.lodgify.net/agiles/86-18/87-784?issue=TE-1405. I'll make a ticket to deduplicate this in tech debt. Is that what you mean?

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
```diff
@@ -34,11 +35,15 @@ export const getModalContentMarkup = (
...
   isUserOnMobile
 ) => (
-  <Modal.Content>
+  <Modal.Content className="room-type">
     <Heading>{name}</Heading>
     {!!ratingNumber && <Rating ratingNumber={ratingNumber} />}
     <Divider size="small" />
-    <Slideshow additionalClass="no-shadow" images={slideShowImages} />
+    <Slideshow
+      additionalClass="no-shadow"
```
marcdavi-es: I know this isn't in scope but does this do anything?

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.spec.js
```diff
@@ -72,4 +71,22 @@ describe('getModalContentMarkup', () => {
       expect(wrapper).toMatchSnapshot();
     });
   });
+
+  describe('`props.slideShowImages.length` > 1', () => {
```
marcdavi-es: describe('if `props.slideShowImages.length` > 1', () => {

src/styles/semantic/themes/livingstone/views/card.overrides
Jonnygwi: Spoke with Julio about this issue. So we going to take a different approach as we hadn't thought about this case when defining the solution. Good thinking! 👍

src/styles/semantic/themes/livingstone/views/card.variables
Jonnygwi: Ahhh I get it. Sure. My mistake

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.spec.js
marcdavi-es:
```suggestion
  describe('if `props.slideShowImages.length` > 1', () => {
```

src/styles/semantic/themes/livingstone/views/card.overrides
stephenkr: Very good, but you don't have to use a ruleset, instead just create a CSS class and extend it here.

src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
Jonnygwi: A prop has been exposed called `hasShadow` which applies this class. You can find it here https://lodgify.github.io/lodgify-ui/#/Media/Slideshow

marcdavi-es: So this should be `hasShadow={false}` ?

src/styles/semantic/themes/livingstone/views/card.overrides
Jonnygwi: https://youtrack.lodgify.net/agiles/86-18/87-784?issue=TE-1502

PR #447, opened 2018-11-22
TE-1460: Reduce the PropertyPageHero height

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1460)

Reduce the height of the `PropertyPageHero` to accommodate the fixed search bar

### Any other notes
- Updated the `Hero` and the `FullBleed` to support a prop called `bottomOffset`. The purpose of `bottomOffset` is to reduce the height of the `FullBleed`.
- Relies on CSS variables 

### Preview 
- The height of the hero has been reduced by `85px`
screen shot 2018-11-22 at 17 15 30

### DIFF
--- a/src/components/collections/Hero/Readme.md
+++ b/src/components/collections/Hero/Readme.md
@@ -2,7 +2,7 @@
 const logoSrc = require('./mock-data/livingstoneLogo.png');
 const { backgroundImageUrl, navigationItems } = require('./mock-data/mock-data');
 
-<Hero     
+<Hero
   backgroundImageUrl={backgroundImageUrl}
   headerLogoSrc={logoSrc}
   headerLogoText="Livingstone Cottage"
@@ -19,7 +19,7 @@ const { backgroundImageUrl, navigationItems } = require('./mock-data/mock-data')
 const logoSrc = require('./mock-data/livingstoneLogo.png');
 const { backgroundImageUrl, navigationItems } = require('./mock-data/mock-data');
 
-<Hero     
+<Hero
   activeNavigationItemIndex={0}
   backgroundImageUrl={backgroundImageUrl}
   headerLogoSrc={logoSrc}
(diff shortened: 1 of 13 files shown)


src/components/collections/Hero/component.js
```diff
@@ -129,6 +132,12 @@ Component.propTypes = {
   headerSearchBarModalHeadingText: PropTypes.string,
   /** The Search Button the Search Bar modal displays. */
   headerSearchBarSearchButton: PropTypes.node,
+  /**
+   * Used internally by `PropertyPageHero` so ignored in the styleguide.
+   * This prop is used to reduce the height of the Hero for the Search bar.
+   * @ignore
+   */
+  isFixedSearchBarDisplayed: PropTypes.bool,
```
marcdavi-es: I don't think this is a good API design. It suggests that components 'know' what is going independently of them.

An alternative approach would be to expose offset props on `FullBleed` e.g. `bottomOffset` which modifies the default bleed by the number of pixels passed?

We can then re-expose `bottomOffset` in `PropertyPageHero` (and whatever else happens to need it).

marcdavi-es: > I'd like to keep the height of the PropertyPageSearchBar in one location.

Why?

stephenkr: On review, I will move forward and implement the `bottomOffset` you suggested, its a good approach in keeping the components non-reliant of each other. Plus it would reduce the amount of complexity around the implementation of the search bar height.

Updating now.

stephenkr: I have updated the change with the approach suggested above.

Resolving this conversation.

src/components/media/FullBleed/Readme.md
```diff
@@ -13,3 +13,12 @@
   hasGradient
 />
 ```
+
+#### With bottom offset applied
```
marcdavi-es:
```suggestion
#### With bottom offset
```

src/components/collections/Hero/Readme.md
```diff
@@ -30,44 +30,43 @@ const { backgroundImageUrl, navigationItems } = require('./mock-data/mock-data')
 />
 ```
 
-### Content
+#### With bottom offset applied
```
marcdavi-es:
```suggestion
#### With bottom offset
```

stephenkr: Very good, fixing now.

PR #448, opened 2018-11-23
TE-1426: In the Form, handle the required validation onChange

stephenkr:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1426)

Applies the `isRequired` error validation to the `handleInputChange`.

### Any other notes
- Also updated the Form documentation to have an empty string option for the `Dropdown` to demonstrate the required state.

### Why this was occurring?
- This bug is specific to the Dropdown.
- We only rely on the blur event to handle the required state, and the change event is only used to handle the isValid state.
- So when a value was changed in the `Dropdown`, the change event would override the state to set the field as valid without taking the required state into account.
- Once the user interacted with the Dropdown again without changing the value, the blur event would then apply the required state if the value was an empty string.

### Before
kapture 2018-11-23 at 11 57 21

### After
kapture 2018-11-23 at 11 58 51

### DIFF
--- a/src/components/collections/Form/component.js
+++ b/src/components/collections/Form/component.js
@@ -8,7 +8,7 @@ import { Heading } from 'typography/Heading';
 import { Link } from 'elements/Link';
 
 import { getValidationWithDefaults } from './utils/getValidationWithDefaults';
-import { getIsRequiredError } from './utils/getIsRequiredError';
+import { getIsRequiredErrorThenSetState } from './utils/getIsRequiredErrorThenSetState';
 import { getIsValidError } from './utils/getIsValidError';
 import { getEmptyRequiredInputs } from './utils/getEmptyRequiredInputs';
 import { getFormChild } from './utils/getFormChild';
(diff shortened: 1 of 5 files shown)


src/components/collections/Form/component.js
```diff
@@ -30,9 +30,24 @@ export class Component extends PureComponent {
...
+
+    if (hasError) {
+      this.setState({
+        [name]: {
+          error: validation.isRequiredMessage,
+        },
+      });
+
+      return;
+    }
```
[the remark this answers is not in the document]
stephenkr: Yes, good point.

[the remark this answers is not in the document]
stephenkr: I have extracted the shared logic out to a utility. So marking as resolved.

src/components/collections/Form/component.js
```diff
@@ -22,14 +22,26 @@ export class Component extends PureComponent {
...
-
-    hasError &&
-      this.setState({ [name]: { error: validation.isRequiredMessage } });
+    getIsRequiredErrorThenSetState(
+      this.setState.bind(this),
+      this.props.validation[name],
+      name,
+      this.state[name]
+    );
   };
```
Jonnygwi: This could just be 
```
handleInputBlur = name =>
  getIsRequiredErrorThenSetState(
    this.setState.bind(this),
    this.props.validation[name],
    name,
    this.state[name]
  );
```

stephenkr: Fixing now.

stephenkr: The PR has been updated and I have tested to ensure no unexpected behavior has been introduced.

PR #451, opened 2018-11-28
TE-1402 Add autocomplete to forms and inputs

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1402)

Adds configurable autocomplete functionality to forms and inputs by exposing `props.autoComplete`.

### Any other notes
This includes some hardwired autoComplete configuration in form-consuming components like `CallMeBack` and others.

kapture 2018-11-28 at 11 44 27

### DIFF
--- a/src/components/collections/Form/component.js
+++ b/src/components/collections/Form/component.js
@@ -70,6 +70,7 @@ export class Component extends PureComponent {
   render = () => {
     const {
       actionLink,
+      autoComplete,
       children,
       errorMessage,
       headingText,
@@ -86,6 +87,7 @@ export class Component extends PureComponent {
         )}
         <Card.Content>
           <Form
+            autoComplete={autoComplete}
             error={!!errorMessage}
             onSubmit={this.handleSubmit}
             success={!!successMessage}
@@ -114,6 +116,7 @@ export class Component extends PureComponent {
 Component.displayName = 'Form';
 
 Component.defaultProps = {
+  autoComplete: null,
   errorMessage: '',
   headingText: null,
   onSubmit: Function.prototype,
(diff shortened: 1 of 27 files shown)


src/components/collections/Form/component.js
```diff
@@ -114,6 +116,7 @@ export class Component extends PureComponent {
 Component.displayName = 'Form';
 
 Component.defaultProps = {
+  autoComplete: null,
```
red-runner: @marcdavi-es Are setting this to `null` to prevent the browser from forcing `autoComplete` ?

> In some cases, the browser will continue suggesting autocompletion values even if the autocomplete attribute is set to off. This unexpected behavior can be quite puzzling for developers. The trick to really enforcing non-autocompletion is to assign an invalid value. [Source](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion)

marcdavi-es: No - it's just a safe 'no value' default.

In the default case, there will be no `autocomplete` attribute on the HTML `<form />` and the browser will do whatever it does.

It's up to the consumer of `Form` to either use this default behaviour or to explicit set 'on' or 'off'.

src/components/collections/Form/component.js
```diff
@@ -131,6 +134,8 @@ Component.propTypes = {
     /** The visible text for the secondary call to action */
     text: PropTypes.node.isRequired,
   }),
+  /** Can inputs be completed automatically by the browser. See [MDN docs for `<form />` for more](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). */
+  autoComplete: PropTypes.oneOf(['on', 'off']),
```
Jonnygwi: Shouldn't we expose it as a boolean called `hasAutoComplete` and adapt the boolean to match these 'on/off' values when passing it through to the form?

PR #452, opened 2018-11-28
TE-1383 Implement ShowOn component

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1383)

Implements the `ShowOn` component to replace the old `ShowOnDesktop` and `ShowOnMobile` components. The new component hides the wrapped children by default and only shows it on de the devices selected by the user. 

### Any other notes
* The added viewports and breakpoints respect the legacy Breakpoints used across lodgify-ui. 

showon_component


src/components/layout/ShowOn/component.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import getClassNames from 'classnames';
+
+/**
+ * ShowOn wraps child components and shows only the selected ones,
+ * the rest remain hidden.
```
marcdavi-es: This can be clearer.

```
Shows child components on specified screen sizes. Hides child components otherwise.
```

src/components/layout/ShowOn/component.js
```diff
@@ -0,0 +1,66 @@
...
+  parentProps,
+  tablet,
+  widescreen,
+}) =>
+  React.createElement(
+    parent,
+    {
+      ...parentProps,
+      className: getClassNames(parentProps.className, {
+        'hide-on-all-devices': true,
```
marcdavi-es: I don't think we need this? We can just do 

```css
.show-on-mobile,
.show-on-tablet,
.show-on-desktop,
.show-on-widescreen {
  display: none !important;
}

...media queries follow and take precedence by cascade
```

red-runner: @marcdavi-es The problem with this is that, if the user selects no classes, implying that sets all the devices as false, the component will show in all devices. Then:
```
Shows child components on specified screen sizes. Hides child components otherwise.
```
will be false. Do we want to care about this or not?

marcdavi-es: Oh I see. Yeah fair enough. Let's go with how you've done it.

One small point on `getClassNames`...

```js
getClassNames(parentProps.className, 'hide-on-all-devices', {
   'show-on-mobile': mobile,
   ...
```

See API at https://github.com/JedWatson/classnames

PR #468, opened 2018-12-13
TE-1436 Availability i18n

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1436)

Exposes string props for the Availability component.

### Any other notes
image

kapture 2018-12-13 at 13 25 56

#### Availability
image

#### Contact
image

#### SearchBar
image

#### HomepageHero
image

#### Header
image

#### PropertyPageSearchBar
image


src/components/inputs/DateRangePicker/component.spec.js
```diff
@@ -42,91 +33,6 @@ describe('<DateRangePicker />', () => {
...
-        customInputIcon: <Icon name={ICON_NAMES.CALENDAR} />,
-        daySize: 52,
-        hideKeyboardShortcutsPanel: true,
-        navNext: <Icon name={ICON_NAMES.ARROW_RIGHT} />,
-        navPrev: <Icon name={ICON_NAMES.ARROW_LEFT} />,
-        numberOfMonths: ReactDatesDateRangePicker.defaultProps.numberOfMonths,
-      });
-    });
-  });
-
```
marcdavi-es: I don't think this is a good case for switching to snapshot.
- the `shallow` snapshot just gives us a Semantic UI `Responsive` component with only a subset of the props we need to make the `DateRangePicker` work.
- if we switch to `mount` I think we'll get an insanely deep snapshot with thousands and thousands of lines, which isn't necessarly helpful either.

I think either we stick to the old way of testing in this case or have a snapshot of the shallow component structure but one level further down.

marcdavi-es: Actually in other places we have pretty long snaps.

In that case I don't mind if we just use `mount` here....

Jonnygwi: As discussed in Slack this will revert to the manual tree comparison as the snapshot is huge.

PR #473, opened 2018-12-20
TE-1576 Hide navigation items if they exceed 900px on desktop

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1576)

Hides the `Header` if  the navigation items exceed 900px on computer and widescreen screens; and, the `Header` will display a mobile-style burger menu in all viewports.

### Any other notes
large_navigation_items


src/components/collections/Header/utils/getStandardMenuMarkup.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+export const getStandardMenuMarkup = (
+  {
+    /* eslint-disable react/prop-types */
+    activeNavigationItemIndex,
+    navigationItems,
+    primaryCTA,
+    /* eslint-enable react/prop-types */
+  },
+  isStandardMenuShowingOnComputer
```
marcdavi-es: Do we need this? 

I think it's always true. https://github.com/lodgify/lodgify-ui/pull/473/files#diff-ed07c99ab29b1b82da1000fb02576851R40

If that's right, we don't need to change any logic in this file, just the name.

src/components/collections/Header/Readme.md
```diff
@@ -153,3 +153,52 @@ const logoSrc = require('./mock-data/livingstoneLogo.png');
   primaryCTA={{ onClick: console.log, text: 'Book now'}}
 />
 ```
+
+#### Large Navigation items
```
marcdavi-es: Maybe `Wide navigation items` is better?

src/components/collections/Header/component.js
```diff
@@ -27,8 +28,18 @@ export const Component = props => (
...
         props.logoSrcSet
       )}
-      {getMobileMenuMarkup(props)}
-      {getDesktopMenuMarkup(props)}
+      {areNavigationItemsExceedingHeaderMaxWidth(
+        HEADER_MAX_WIDTH,
+        props.navigationItems,
+        props.primaryCTA
+      ) ? (
+        <Fragment>{getHiddenMenuMarkup(props, true)}</Fragment>
```
marcdavi-es: I don't think we need the `Fragment` wrapper here?

src/components/collections/Header/component.js
```diff
@@ -27,8 +28,18 @@ export const Component = props => (
...
+      ) : (
+        <Fragment>
+          {getHiddenMenuMarkup(props, false)}
+          {getStandardMenuMarkup(props, true)}
```
marcdavi-es: See my comment on `getStandardMenuMarkup`. If you agree, we can remove the second argument here.

src/components/collections/Header/mock-data/navigationItems.js
```diff
@@ -21,4 +21,37 @@ export const navigationItems = [
       },
     ],
   },
+  { text: 'Contact Us', href: '/' },
+  { text: 'What to do', href: '/' },
+];
+
+export const LargeNavigationItems = [
```
marcdavi-es:
```suggestion
export const largeNavigationItems = [
```

src/components/collections/Header/utils/getStandardMenuMarkup.js
red-runner: No we don't. It's always true 👍

src/components/collections/Header/utils/areNavigationItemsExceedingHeaderMaxWidth.js
```diff
@@ -0,0 +1,20 @@
...
+import { getPrimaryCTAWidth } from './getPrimaryCTAWidth';
+
+/**
+ * @param  {number} headerMaxWidth
+ * @param  {Array}  navigationItems
+ * @param  {Object} primaryCTA
+ * @return {boolean}
+ */
+export const areNavigationItemsExceedingHeaderMaxWidth = (
+  headerMaxWidth,
```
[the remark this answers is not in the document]
red-runner: @Jonnygwi I thought that there was no harm in making this constant explicit in the component. This way it's very easy to know which factors we are taking into account inside `areNavigationItemsExceedingHeaderMaxWidth`.  What do you think?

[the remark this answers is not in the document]
Jonnygwi: Yeah I hear what you are saying @red-runner . The name of the function also has `HeaderMaxWidth` within it so it's clear from that what the function is depending on.

PR #477, opened 2019-01-10
TE-1622 DateRangePicker and SingleDatePicker; Open as modal on viewports with height < 850px

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1622)

Implements the `withPortal` prop for both `DateRangePicker` and `SingleDatePicker` components on viewports with height < 850px 

### Additional Notes
* The modal display will require some CSS adjustments; thought It would be better to handle them separately. 

### Behaviour
datepicker


red-runner:
@marcdavi-es Will do👌

red-runner:
@marcdavi-es css commit added 👌

src/components/inputs/SingleDatePicker/component.js
```diff
@@ -18,6 +20,11 @@ export class Component extends PureComponent {
   state = {
     date: null,
     isFocused: null,
+    windowHeight: getWindowHeight(),
+  };
+
+  componentDidMount = () => {
+    global.addEventListener('resize', this.handleHeightChange);
```
Jonnygwi: We should probably add a debounce to this otherwise `this.handleHeightChange` will be called an unnecessary amount of times. Have a look at this. https://lodash.com/docs/4.17.11#debounce

red-runner:
@Jonnygwi This was solved with the css commit that was introduced at the end, following @marcdavi-es suggestion to include it. I'll upload the final view so you can see it.

PR #484, opened 2019-01-17
TE-1694 Kill production

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1694)

Keeps branch naming consistent across projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,13 @@ deploy:
     keep-history: true
     local-dir: './styleguide/dist'
     on:
-      branch: production
+      branch: master
     provider: pages
     skip_cleanup: true
   # Release npm
   - provider: script
     on:
-      branch: production
+      branch: master
     skip_cleanup: true
     script:
-      - npx semantic-release --branch production
+      - npx semantic-release
--- a/Readme.md
+++ b/Readme.md
@@ -66,4 +66,4 @@ import '@lodgify/ui/lib/styles/lodgify-ui.css';
 
 ### Contributing
 
-See  [contributing](https://github.com/lodgify/lodgify-ui/blob/production/docs/CONTRIBUTING.md).
+See  [contributing](https://github.com/lodgify/lodgify-ui/blob/master/docs/CONTRIBUTING.md).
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
   "scripts": {
     "build": "npm-run-all -p build:less build:js -s rm:build-orphan",
     "build:less": "webpack --config $npm_package_config_less_build_config",
-    "build:js": "BABEL_ENV=production babel src/ -d lib/",
+    "build:js": "BABEL_ENV=master babel src/ -d lib/",
```
marcdavi-es: This should be left as `production`. This is part of a different convention. 

https://new.babeljs.io/docs/en/next/babelrc.html#env-environment-option

.babelrc
```diff
@@ -1,6 +1,6 @@
 {
   "env": {
-    "production": {
+    "master": {
```
marcdavi-es: See https://github.com/lodgify/lodgify-ui/pull/484/commits/c4992e4a31606911643e41e2ee25c9987ea3e5f4#r248760280

.travis.yml
```diff
@@ -19,13 +19,13 @@ deploy:
...
     skip_cleanup: true
   # Release npm
   - provider: script
     on:
-      branch: production
+      branch: master
     skip_cleanup: true
     script:
-      - npx semantic-release --branch production
+      - npx semantic-release --branch master
```
marcdavi-es: Check the travis and semantic release docs - I think all of these things can be removed since the tools default to `master`

Jonnygwi: You are just refering to the `--branch master` right?

marcdavi-es: Ah - now I see the confusion.

By 'All of these things' I meant:
- `--branch master` on this line
- `on: \ branch: master` on lines 21/22 and 27/28

I think all of these can be removed.

Jonnygwi: Ahhhh got you now.. Great. I'll have a look at what the default values are and then make the appropriate changes.

Jonnygwi: @marcdavi-es the `on:` is used so that the deploy step isn't ran whenever a merge to any branch is made. In this case I think we need it.

PR #489, opened 2019-01-18
TE-1686 HostProfile Description is not required

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1686)

Makes the `description` prop not required in `HostProfile`; then, implements `getDescription` util to handle `undefined`

### Behaviour
hostprofile

### DIFF
--- a/src/components/property-page-widgets/HostProfile/utils/getDescription.js
+++ b/src/components/property-page-widgets/HostProfile/utils/getDescription.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import { size } from 'lodash';
+
+import { GridColumn } from 'layout/GridColumn';
+import { Paragraph } from 'typography/Paragraph';
+
+/**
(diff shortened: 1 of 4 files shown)


src/components/property-page-widgets/HostProfile/utils/getDescription.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { size } from 'lodash';
+
+import { GridColumn } from 'layout/GridColumn';
+import { Paragraph } from 'typography/Paragraph';
+
+/**
+ * @param {string} description
+ * @return {Object}
```
Jonnygwi: This should be

```suggestion
 * @return {Object|null}
```

You'll also have to define null so the top line of the JSdoc should be 

` * @typedef {Object} null`

Check `Dropdown/utils/getDefaultValue.js`

PR #492, opened 2019-01-21
TE-1647 Hide gallery with less than two images

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1647)

Hides the `PropertyPageHero` gallery when there are fewer than two images

### Behaviour
screenshot 2019-01-21 at 12 32 50

### DIFF
--- a/src/components/property-page-widgets/PropertyPageHero/component.js
+++ b/src/components/property-page-widgets/PropertyPageHero/component.js
@@ -1,19 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { upperCase } from 'lodash';
 
 import { withResponsive } from 'utils/with-responsive';
 import { Hero } from 'collections/Hero';
-import { HorizontalGutters } from 'layout/HorizontalGutters';
-import { FlexContainer } from 'layout/FlexContainer';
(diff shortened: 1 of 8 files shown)


src/components/property-page-widgets/PropertyPageHero/component.js
```diff
@@ -103,7 +83,7 @@ Component.defaultProps = {
   propertyName: null,
   ratingNumber: null,
   searchBarGetIsDayBlocked: undefined,
-  searchBarModalHeadingText: undefined,
+  searchBarModalHeadingText: '',
```
red-runner: @marcdavi-es @Jonnygwi I was getting a warning about the `Heading` component not having any children. This solved the warning, although I'm not sure if it's the right solution. Have a look.

marcdavi-es: In `HomepageHero` we have

```js
searchBarModalHeadingText: CHECK_OUR_AVAILABILITY,
```

I think that works OK here too?

red-runner: Yeah, this will work better 🎉

PR #501, opened 2019-01-24
 TE-1690 Footer should be fixed to the bottom of the page

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1690)

Implements the new `Viewport` component; then implements the missing footer styles requested in the card.

### Any other notes
- Another PR will need to be opened in templates-ssr to consume `Viewport`

### Component behaviour
height


src/components/layout/Viewport/component.js
```diff
@@ -0,0 +1,32 @@
...
+// eslint-disable-next-line jsdoc/require-jsdoc
+export const Component = ({ children }) => (
+  <div
+    style={{
+      display: 'flex',
+      flexGrow: '1',
+      flexDirection: 'column',
+      minHeight: '100vh',
+      position: 'relative',
+      paddingBottom: FOOTER_TOTAL_HEIGHT,
```
marcdavi-es: I'm not sure I understand why we need this?

red-runner:
Switched the approach in order to avoid absolute positioning for the footer.

src/styles/semantic/themes/livingstone/globals/site.overrides
```diff
@@ -45,6 +45,15 @@ h4.ui.header {
...
 
+/*--------------
+  Page Footer
+---------------*/
+
+footer {
+  margin-top: auto;
+  position: relative;
+}
+
```
marcdavi-es: Styles for footer can live with the others in `menu.overrides`

src/components/layout/Viewport/component.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * A container for accomodating children in full viewport height using
+ * [flexbox method](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox).
+ * Fills the full width of its parent and the full height of the viewport
```
marcdavi-es: This is a bit of a mouthful. 

I would lose lines 5 and 6 and just have 

```
* A viewport fills the width of its parent element and the inner height of the browser window
```

PR #503, opened 2019-01-25
TE-1715 Hide Check-in and Check-out from Rules whenever there is no info to show

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1715)

Implements a new util to cover all the cases exposed in the ticket; then implements them in the `Rules` component. 

### Behaviour

behaviour


src/components/property-page-widgets/Rules/utils/getCheckInAndCheckOutMarkup.js
```diff
@@ -0,0 +1,66 @@
...
+        <GridColumn computer={9} tablet={7}>
+          <Icon
+            labelText={getLabelAndValueString(checkInTimeLabel, checkInTime)}
+            name={ICON_NAMES.CHECK_IN}
+          />
+        </GridColumn>
+      );
+    default:
+      return null;
+  }
```
marcdavi-es: This can be simplified.

`checkinTimeLabel` and `checkOutTimeLabel` will always be defined, as they have a `defaultProps` definition. We are concerned only with the `checkInTime` and `checkOutTime` values.

So... something like this is enough

```js
(!!checkInTime || !!checkOutTime) && (
  <GridColumn computer={9} tablet={7}>
    {
      !!checkInTime && (
        <Icon
          labelText={getLabelAndValueString(checkInTimeLabel, checkInTime)}
          name={ICON_NAMES.CHECK_IN}
        />
        <Divider />
      )
    }
    {
      !checkOutTime && (
        <Icon
          labelText={getLabelAndValueString(checkOutTimeLabel, checkOutTime)}
          name={ICON_NAMES.CHECK_OUT}
        />
      )
    }
  </GridColumn>
)
```

PR #511, opened 2019-01-29
TE-1702 Fix `Thumbnail`

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1702)

Fixes `Thumbnail` so that:
- the image fills the thumbnail (provided that the image passed has sufficient width)
- the image is centered in the thumbnail

### Any other notes
I changed `HostProfile` to consume `Thumbnail` as it was distorting the avatar image.

### DIFF
--- a/src/components/media/Thumbnail/component.js
+++ b/src/components/media/Thumbnail/component.js
@@ -64,7 +64,7 @@ Component.propTypes = {
    * @ignore
    */
   className: PropTypes.string,
-  /** Does the thumbnail have margin to separate it visually from siblings. */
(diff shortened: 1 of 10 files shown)


src/components/media/Thumbnail/component.js
```diff
@@ -61,6 +64,8 @@ Component.propTypes = {
    * @ignore
    */
   className: PropTypes.string,
+  /** Does the thumbnail have margin to separate it visually from siblings. */
```
Jonnygwi:
```suggestion
  /** Does the thumbnail have a margin to separate it visually from siblings. */
```

PR #516, opened 2019-01-30
TE-1685 Fix Desktop Menu Modal layout

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1685)

Fixes the regression caused by the merge of the mentioned ticket.

### Current visible menu
before-logo-header

### Current hidden menu
hidden-logo-before

### Current Modal
modal-overflow-before

### Visible menu after fix
visible-menu-after

### Hidden menu after fix
hidden-menu-before

### Modal after fix
modal-after

### DIFF
--- a/src/components/collections/Header/__snapshots__/component.spec.js.snap
+++ b/src/components/collections/Header/__snapshots__/component.spec.js.snap
@@ -69,12 +69,14 @@ exports[`<Header /> by default should render the right structure 1`] = `
           <div
             className="ui borderless ui text container menu"
           >
-            <Item
+            <MenuItem
               href="/"
+              link={true}
             >
               <a
-                className="item"
+                className="link item"
                 href="/"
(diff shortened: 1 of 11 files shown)


src/components/collections/Header/utils/getHiddenMenuMarkup.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Menu, Accordion } from 'semantic-ui-react';
 import { size } from 'lodash';
+import { Item } from 'semantic-ui-react';
```
marcdavi-es: Looks like this line is adding not removing?

red-runner: @marcdavi-es Correct, this commit can be removed entirely...

PR #524, opened 2019-02-06
TE-1781 Dropdown: enable arrow key navigation of options

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1781)

`Dropdown` enable arrow key navigation of options

dropdown-after


src/components/inputs/Dropdown/utils/getIsOpenAfterChange.spec.js
```diff
@@ -0,0 +1,33 @@
...
+      testCases.forEach(testCase => {
+        const actual = getIsOpenAfterChange(testCase);
+
+        expect(actual).toBe(true);
+      });
+    });
+  });
+
+  describe('if `key` is anything else', () => {
+    it('should return `true`', () => {
```
red-runner: Could this be `false` instead of `true` ?

PR #526, opened 2019-02-06
TE-1818 Footer can accept no props.languageOptions and props.currencyOptions

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1818)

Removes the language and currency `Submenu` from the `Footer` when `languageOptions` and `currencyOptions` aren't passed; then, updates the mentioned cases in the docs. 

### Behaviour
screenshot 2019-02-06 at 17 34 25

### DIFF
--- a/src/components/collections/Footer/component.js
+++ b/src/components/collections/Footer/component.js
@@ -54,26 +54,30 @@ export const Component = ({
     </div>
     <div className="bottom-navigation">
       <HorizontalGutters as={Menu} borderless inverted stackable>
-        <Menu.Item>
-          <Submenu
-            items={languageOptions}
-            name="language"
-            onChange={onChangeLanguage}
-            value={languageValue}
-            willOpenAbove
-          />
-        </Menu.Item>
-        <Menu.Item>
-          <Submenu
-            isSearchable
-            items={currencyOptions}
-            name="currency"
-            noResultsText={currencyNoResultsText}
-            onChange={onChangeCurrency}
-            value={currencyValue}
-            willOpenAbove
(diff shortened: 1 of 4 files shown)


src/components/collections/Footer/component.js
```diff
@@ -54,26 +54,30 @@ export const Component = ({
...
-            isSearchable
-            items={currencyOptions}
-            name="currency"
-            noResultsText={currencyNoResultsText}
-            onChange={onChangeCurrency}
-            value={currencyValue}
-            willOpenAbove
-          />
-        </Menu.Item>
+        {size(languageOptions) > 0 && (
```
marcdavi-es:
```suggestion
        {size(languageOptions) > 1 && (
```

right?

src/components/collections/Footer/component.js
```diff
@@ -54,26 +54,30 @@ export const Component = ({
...
+            <Submenu
+              items={languageOptions}
+              name="language"
+              onChange={onChangeLanguage}
+              value={languageValue}
+              willOpenAbove
+            />
+          </Menu.Item>
+        )}
+        {size(currencyOptions) > 0 && (
```
marcdavi-es:
```suggestion
        {size(languageOptions) > 1 && (
```

Jonnygwi: I think Marc is right. If there is only one option then there isn't any other to choose between :)

Jonnygwi:
Removing myself as an assignee as I agree with Marc's requested changes.

PR #528, opened 2019-02-07
TE-1754 Fix getHasObjectProperty regression

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1754)

Fixes the regression caused by #520. The main problem being that `getHasObjectProperty` util wasn't handling `undefined`; causing integration to break.  

### Any other notes
screenshot 2019-02-07 at 12 42 19


src/utils/get-has-object-property/getHasObjectProperty.js
```diff
@@ -4,4 +4,4 @@
  * @return {boolean}
  */
 export const getHasObjectProperty = (options, property) =>
-  options.some(option => option.hasOwnProperty(property));
+  !!options ? options.some(option => option.hasOwnProperty(property)) : false;
```
marcdavi-es:
```suggestion
options && options.some(option => option.hasOwnProperty(property));
```

is cleaner

PR #535, opened 2019-02-11
TE-1883 Divider standardise margins

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1883)

Dividers with a size of large have decreased in size and Dividers of size huge have replaced the previous large size.

### Any other notes
- The css selector has been changed because `is-size-small` class wasn't getting any css applied to it.
- Size `large` has been changed to `@20px` and `huge` was added with `@40px`

### DIFF
--- a/src/components/elements/Divider/component.js
+++ b/src/components/elements/Divider/component.js
@@ -13,6 +13,7 @@ export const Component = ({ className, hasLine, size }) => (
   <Divider
     className={getClassNames(className, {
       'is-size-small': size === 'small',
+      'is-size-huge': size === 'huge',
     })}
     hidden={!hasLine}
     section={getIsSizeLarge(size)}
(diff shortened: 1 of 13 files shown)


src/components/elements/Divider/component.js
```diff
@@ -37,5 +38,5 @@ Component.propTypes = {
   /** Does the divider have a visible line. */
   hasLine: PropTypes.bool,
   /** The size of the divider. */
-  size: PropTypes.oneOf(['small', 'medium', 'large']),
+  size: PropTypes.oneOf(['small', 'medium', 'large', 'huge']),
```
marcdavi-es: What's the difference between `'medium'` and `'large'`?

Jonnygwi: | size name | size |
|-----------|-----|
| small    |  0.57142857rem |
| medium | 1rem |
| large | 1.42857143rem |
| huge | 2.85714286rem |

PR #544, opened 2019-02-14
TE-1880 Contact DateRangePicker

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1880)

DateRangePicker is on its own row and email and guests inputs are on the same row.

### Any other notes
image

### DIFF
--- a/src/components/general-widgets/Contact/component.js
+++ b/src/components/general-widgets/Contact/component.js
@@ -65,13 +65,11 @@ export const Component = ({
       <TextInput autoComplete="name" label={nameInputLabel} name="name" />
       <PhoneInput autoComplete="tel" label={phoneInputLabel} name="phone" />
     </InputGroup>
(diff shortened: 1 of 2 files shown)


src/components/general-widgets/Contact/component.js
```diff
@@ -65,22 +65,21 @@ export const Component = ({
...
-      <NumberInput
-        label={guestsInputLabel}
-        min={1}
-        name="guests"
-        width="four"
+      <TextInput
+        autoComplete="email"
+        label={emailInputLabel}
+        name="email"
+        width="five"
```
marcdavi-es: How come this isn't `"eight"` and the `NumberInput` `"four"` ?

Jonnygwi: Hmmm strange. This as "five" has the same effect as having email as "eight" and number input as "four".

The eight and four options is clearer so I'll make that change.

PR #559, opened 2019-03-04
TE-1930 Pictures: expose props.numberOfThumbnails

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1930)

`Pictures`: expose `props.numberOfThumbnails`

screenshot 2019-03-04 at 11 45 57


src/utils/get-first-n-items/getFirstNItems.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * @param  {number}   number
+ * @return {Function}
+ */
+export const getFirstNItems = number => items => items.slice(0, number);
```
Jonnygwi: Why have you chosen to follow this currying pattern rather than just have two arguments?

marcdavi-es: I don't know

PR #576, opened 2019-03-11
TE-1982 Footer hide container

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1982)
[Related YouTrack issue](https://youtrack.lodgify.net/agiles/86-18/87-1073?issue=TE-1980)

- Hide the navigation container if there are no navigation items.
- Aligns the searchable dropdown in the footer

### Any other notes
image
Screenshot 2019-03-11 at 15 49 12

### DIFF
--- a/src/components/collections/Footer/component.js
+++ b/src/components/collections/Footer/component.js
@@ -34,24 +34,26 @@ export const Component = ({
   socialMediaLinks,
 }) => (
   <footer>
-    <div className="top-navigation">
-      <HorizontalGutters as={Menu} borderless inverted stackable>
-        {getAreNavigationItemsGrouped(navigationItems)
-          ? getGroupedNavigationItems(navigationItems).map(
-              ({ text, subItems }, index) => (
-                <Menu.Item
-                  key={buildKeyFromStrings(text || subItems[0].text, index)}
-                >
-                  <Menu.Menu>
-                    {text && <Menu.Header>{toUpper(text)}</Menu.Header>}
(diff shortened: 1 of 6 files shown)


src/components/collections/Footer/component.js
```diff
@@ -122,6 +124,7 @@ Component.defaultProps = {
   currencyValue: null,
   languageOptions: [],
   languageValue: null,
+  navigationItems: [],
```
marcdavi-es:
```suggestion
  navigationItems: null,
```
I don't think we need the array for anything?

src/components/collections/Footer/component.js
```diff
@@ -34,24 +34,26 @@ export const Component = ({
...
-              ({ text, subItems }, index) => (
-                <Menu.Item
-                  key={buildKeyFromStrings(text || subItems[0].text, index)}
-                >
-                  <Menu.Menu>
-                    {text && <Menu.Header>{toUpper(text)}</Menu.Header>}
-                    {subItems.map(getMenuItemMarkup)}
-                  </Menu.Menu>
-                </Menu.Item>
+    {size(navigationItems) > 0 && (
```
marcdavi-es:
```suggestion
    {size(navigationItems) && (
```
is enough, no?

Jonnygwi: The above isn't enough as it returns a `0` but `{!!size(navigationItems) && (` is.. What do you recon? In other places we've done ` > 0` like on the `socialMediaLinks`.

PR #578, opened 2019-03-12
TE-1904 Persist search value with a cookie

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1904)

### What are the 3 things this PR does?

- Removes unnecessary `defaultValue` from `Dropdown`.
- Adds initial values for `SearchBar` state inputs.
- Fixes `HomepageHero` and `PropertySearchBar` guest input PropTypes.

### DIFF
--- a/src/components/general-widgets/HomepageHero/component.js
+++ b/src/components/general-widgets/HomepageHero/component.js
@@ -236,7 +236,11 @@ Component.propTypes = {
    */
   searchBarGetIsDayBlocked: PropTypes.func,
   /** The initial value for the guests input of the search bar. */
-  searchBarGuestsInputInitialValue: PropTypes.string,
+  searchBarGuestsInputInitialValue: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.number,
+    PropTypes.string,
+  ]),
   /** The options which the user can select in the guests fields of the search bar. */
   searchBarGuestsOptions: PropTypes.arrayOf(
     PropTypes.shape({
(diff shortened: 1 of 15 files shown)


src/components/general-widgets/HomepageHero/component.js
```diff
@@ -236,7 +236,11 @@ Component.propTypes = {
    */
   searchBarGetIsDayBlocked: PropTypes.func,
   /** The initial value for the guests input of the search bar. */
-  searchBarGuestsInputInitialValue: PropTypes.string,
+  searchBarGuestsInputInitialValue: PropTypes.oneOfType([
+    PropTypes.bool,
```
Jonnygwi: When is this value a `boolean`?

src/components/general-widgets/SearchBar/component.js
```diff
@@ -20,7 +20,11 @@ import { getSearchBarModal } from './utils/getSearchBarModal';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export class Component extends PureComponent {
-  state = {};
+  state = {
+    dates: { endDate: null, startDate: null },
+    guests: null,
+    location: null,
+  };
```
marcdavi-es: Doesn't this need to be populated with the initial values if they are available? Or does it get those on the first update?

src/components/general-widgets/HomepageHero/component.js
red-runner: @Jonnygwi I am slightly uncertain for this as well. I've checked elsewhere, like in the `Footer` component and I found this: 
```
  /** The current value of the language dropdown. */
  languageValue: PropTypes.oneOfType([
    PropTypes.bool,
    PropTypes.number,
    PropTypes.string,
  ]),
```
What do you think?

PR #581, opened 2019-03-15
TE-2008 Add lazy loading

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2008)

Proposal for lazy loading

### Any other notes
- The debounce was increased to make the lazy loading more clear

Kapture 2019-03-15 at 15 39 45

### DIFF
--- a/src/components/layout/LazyLoader/component.js
+++ b/src/components/layout/LazyLoader/component.js
@@ -1,0 +1,102 @@
+import React, { PureComponent } from 'react';
+import { debounce } from 'lodash';
+import PropTypes from 'prop-types';
+
+import { getIsVisible } from './utils/getIsVisible';
+
+/**
+ * A wrapper that renders components when they are visible.
+ */
+// eslint-disable-next-line jsdoc/require-jsdoc
+export class Component extends PureComponent {
+  state = {
+    shouldRender: false,
+  };
+
+  componentDidMount = () => {
+    this.addEventListeners();
+    this.shouldComponentRender();
+  };
+
+  componentWillUnmount = () => this.removeEventListeners();
+
+  shouldComponentRender = () => {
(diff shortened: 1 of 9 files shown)


src/components/layout/LazyLoader/component.js
```diff
@@ -0,0 +1,102 @@
+import React, { PureComponent } from 'react';
+import { debounce } from 'lodash';
+import PropTypes from 'prop-types';
+
+import { isVisible } from './utils/isVisible';
```
marcdavi-es:
```suggestion
import { getIsVisible } from './utils/getIsVisible';
```

src/components/layout/LazyLoader/component.js
```diff
@@ -0,0 +1,102 @@
...
+            ...componentProps,
+          }
+        : componentProps;
+
+    const componentToRender =
+      lazyProps !== undefined || this.state.shouldRender
+        ? React.cloneElement(lazyComponent, componentToRenderProps)
+        : null;
+
+    return <div ref={this.createComponentRef}>{componentToRender}</div>;
```
marcdavi-es: I wonder if this `div` might break some component structures?

Can we just use a dummy element which does not wrap `componentToRender`? Then maybe we can render the dummy first, use to watch the position relative to the viewport, then when it is visible, swap it out for `componentToRender` ?

Jonnygwi: Hmmmmmm Mmmmmmm.

Jonnygwi: Good thought. I'll give it a try.

Jonnygwi: I'm going to leave it as a div for now and if it effects some component layouts then we can make optimisations then.. As for the moment it's only going to be used in Images and Gallery so it should change anything there.

PR #583, opened 2019-03-22
TE-2083 DateRangePicker: change styles for blocked and unblocked days

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2083)

Implements the requested styles for `DateRangePicker`; blocked and unblocked dates on hover. 

### Additional Notes
The following bug was solved in this PR:

#### Before
Screenshot 2019-03-25 at 16 49 37

#### After
Screenshot 2019-03-25 at 17 38 26

### Current behaviour
daterangepicker

### DIFF
--- a/src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
+++ b/src/styles/semantic/definitions/third-party-components/react-dates-datepicker.less
@@ -144,6 +144,16 @@
       &.CalendarDay__hovered_span {
         background: @pickerDayHoveredBackground;
         color: inherit;
+
+        &.CalendarDay__blocked_calendar {
+          background: @pickerDayHoveredBackground;
+          color: inherit;
+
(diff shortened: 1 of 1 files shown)


red-runner:
> In the example, I don't think 5 April 2017 fits with Case 1 as described in the requirement.
> 
> I think it should be light yellow?

@marcdavi-es I checked out with Julio before opening PR, and he confirmed that this was the correct behaviour. Also, in Zepplin, there's no yellow hover for the blocked dates:

Screenshot 2019-03-22 at 16 31 19

marcdavi-es:
@red-runner 

That's Case 3 in the requirement, when the date is hovered.

Case 1 has the blocked date with light yellow background when it is not hovered...

Screenshot 2019-03-22 at 16 33 52

PR #586, opened 2019-03-22
TE-2029 Hide amenities bar when there's nothing to display

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2029)

Removes the Room Amenities `headingText` when there are no items.

### Any other notes
availableamenities

### DIFF
--- a/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
+++ b/src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
@@ -60,11 +60,13 @@ export const getModalContentMarkup = (
       ))}
     </List>
     <Divider hasLine />
-    <Amenities
-      amenities={amenities}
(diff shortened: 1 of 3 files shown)


src/components/property-page-widgets/RoomType/utils/getModalContentMarkup.js
```diff
@@ -58,7 +58,7 @@ export const getModalContentMarkup = (
     <Divider hasLine />
     <Amenities
       amenities={amenities}
-      headingText="Room Amenities"
+      headingText={size(amenities) > 0 ? 'Room Amenities' : null}
       isStacked={isUserOnMobile}
     />
```
marcdavi-es: We can hide the whole of the `Amenities` component if there are no amenities...

```js
{!!size(amenities) && <Amenities ...
```

PR #605, opened 2019-04-16
TE-1722 Slider components should have a fixed height and background colour before loading

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1772)

Implements two new handlers to determine the height of the slider `onResize` and `onLoad`. Then, implements a `defaultImage` prop to avoid the slideshow to collapse when an image is broken.

### Notes
@marcdavi-es I know we spoke about the `defaultImage`. When an image is broken, the `onImageLoad` event calculates the height based on the small broken image icon, this collapsing the slideshow. I think having a default is the simplest way to solve this. 

### Behaviour
Kapture 2019-04-16 at 12 00 56

### Empty Cache/Hard Reload + Images Broken
* Every image is unique to represent the loading better
Image-error


src/components/media/Slideshow/component.js
```diff
@@ -71,6 +84,8 @@ Component.defaultProps = {
 };
 
 Component.propTypes = {
+  /** An src pointing to a default image if an image fails to load. */
+  defaultImage: PropTypes.string.isRequired,
```
marcdavi-es: 1. I don't think this should be required?
2. I suggest the following doc string and name to match the existing API more closely

```js
      /** URL pointing to the default image to display if an image fails to load. */
      defaultImageUrl: PropTypes.string,
```

src/components/media/Slideshow/component.js
```diff
@@ -20,10 +22,19 @@ export class Component extends PureComponent {
     index: 0,
   };
 
+  componentDidMount = () => {
+    global.window.addEventListener('resize', this.handleResize);
+  };
+
   handleSlide = index => this.setState({ index });
 
+  handleResize = () => setGallerySlidesHeight();
```
marcdavi-es: Debounce this guy.

src/components/media/Slideshow/component.js
```diff
@@ -45,8 +56,10 @@ export class Component extends PureComponent {
           })}
           // Note: styles for the pagination controls
           // live in `styles/semantic/themes/livingstone/collections/menu.*`
+          defaultImage={defaultImage}
           items={adaptImages(images)}
           lazyLoad
+          onImageLoad={this.handleImageLoad}
```
marcdavi-es: Does `onImageLoad` get called only when an image successfully loads

src/components/media/Slideshow/utils/getLoadedImagesHeight.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @return {string}
+ */
+export const getLoadedImagesHeight = () => {
+  const galleryImageWrapper = document.querySelector('.image-gallery-image');
```
marcdavi-es: `global.document.querySelector`

src/components/media/Slideshow/utils/setGallerySlidesHeight.js
```diff
@@ -0,0 +1,11 @@
...
+/**
+ * @return {undefined}
+ */
+export const setGallerySlidesHeight = () => {
+  const imageGalleries = document.querySelectorAll('.image-gallery-slides');
+
+  imageGalleries.forEach(
+    gallery => (gallery.style.height = `${getLoadedImagesHeight()}`)
+  );
+};
```
marcdavi-es: 1. remember to use `global.document`
2. This and `setHeightFromEventTarget` do more or less the same thing. Better to write one function which accepts a string argument `height`. i.e.

```js
export const setImageGallerySlidesElementsHeight = height => {
  const imageGalleries = global.document.querySelectorAll('.image-gallery-slides');

   imageGalleries.forEach(
    gallery => (gallery.style.height = height)
  );
}
```

src/components/media/Slideshow/utils/setGallerySlidesHeight.spec.js
```diff
@@ -0,0 +1,37 @@
+jest.mock('./getLoadedImagesHeight');
+
+import { setGallerySlidesHeight } from './setGallerySlidesHeight';
+import { getLoadedImagesHeight } from './getLoadedImagesHeight';
+
+const testCases = [
```
marcdavi-es: These aren't test cases, they are mock return values.

src/components/media/Slideshow/utils/setGallerySlidesHeight.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  it('should call `document.querySelectorAll` with the right argument', () => {
+    setGallerySlidesHeight();
+
+    expect(document.querySelectorAll).toHaveBeenCalledWith(selector);
+  });
+
+  it('should call `getLoadedImagesHeight` for each test case', () => {
+    setGallerySlidesHeight();
+
+    expect(getLoadedImagesHeight).toHaveBeenNthCalledWith(3);
```
marcdavi-es: This asserts that `getLoadedImagesHeight` is called the third time with no arguments. It is true, but it is not a valid test for the case you state.

```js
    expect(getLoadedImagesHeight).toHaveBeenCalledTimes(3);
```

src/components/media/Slideshow/utils/setGallerySlidesHeight.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
```
marcdavi-es: You're missing a case which tests that each `gallery.style.height` gets set with the right value.

src/components/media/Slideshow/utils/setHeightFromEventTarget.js
```diff
@@ -0,0 +1,11 @@
...
+ * @param {Event} event
+ * @return {string}
+ */
+export const setHeightFromEventTarget = event => {
+  const imageGalleries = document.querySelectorAll('.image-gallery-slides');
+
+  imageGalleries.forEach(
+    gallery => (gallery.style.height = `${event.target.height}px`)
+  );
+};
```
marcdavi-es: See https://github.com/lodgify/lodgify-ui/pull/605/commits/77debee0b46ea805ad5cf233c377740208da58ea#r275763925

src/components/media/Slideshow/component.js
```diff
@@ -45,8 +56,10 @@ export class Component extends PureComponent {
           })}
           // Note: styles for the pagination controls
           // live in `styles/semantic/themes/livingstone/collections/menu.*`
+          defaultImage={defaultImage}
```
marcdavi-es: I'm not sure this is a good solution. If any image is failing, it seems that the default image could fail as well?

src/components/media/Slideshow/component.js
red-runner: @marcdavi-es If we debounce him, the grey background will be seen `onResize`. Can we assume that the user will just do punctual a puntual resize once in a while? If so, then might be acceptable.

src/components/media/Slideshow/component.js
red-runner: @marcdavi-es If the image is broken, `onImageLoad` won't get called. Every time `onImageLoad` gets called, its event callback contains the images that have been loaded when the user goes from one slide to another. The first time though, I think it loads the first 2 images.

src/components/media/Slideshow/utils/getLoadedImagesHeight.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @return {string}
+ */
+export const getLoadedImagesHeight = () => {
+  const galleryImageWrapper = global.document.querySelector(
+    '.image-gallery-image'
+  );
```
marcdavi-es: This might fail if we have other slideshows of different sizes in the same page.

We can target a specific element using refs.

src/components/media/Slideshow/utils/setImageGallerySlidesElementsHeight.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @param {string} height
+ * @return undefined
+ */
+export const setImageGallerySlidesElementsHeight = height => {
+  const imageGalleries = global.document.querySelectorAll(
+    '.image-gallery-slides'
+  );
```
marcdavi-es: Same concern as https://github.com/lodgify/lodgify-ui/pull/605/commits/391ccf43bfab954b3cbb2881b9a761cf2269c1f9#r278586472

src/components/media/Slideshow/utils/setImageGallerySlidesElementsHeight.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @param {string} height
+ * @return undefined
```
marcdavi-es: Shouldn't this be  `* @return {undefined}`

src/components/media/Slideshow/utils/renderBlockPlaceholder.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import { BlockPlaceholder } from 'elements/BlockPlaceholder';
+
+/**
+ * @return Object
```
marcdavi-es: ` * @return {Object}`

PR #610, opened 2019-04-18
TE-2063 input error on blur

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2063)

Form inputs show error state on blur and hides on focus.

### Any other notes
Kapture 2019-04-17 at 15 31 16


src/components/collections/Form/component.js
```diff
@@ -32,32 +31,17 @@ export class Component extends PureComponent {
     }
 
     Object.entries(this.state).forEach(([inputName, inputState]) => {
-      const previousInputState = previousState[inputName] || {};
+      const previousInputState = previousState[inputName];
 
-      if (isEqual(previousInputState, inputState)) return;
-
-      if (previousInputState.value !== inputState.value) {
```
marcdavi-es: In this case we reset the `isValid` and `error` properties to `undefined`.

src/components/collections/Form/component.js
```diff
@@ -32,32 +31,17 @@ export class Component extends PureComponent {
...
-        const validationState = getValidationState(
-          this.props.validation[inputName],
-          inputState.value
-        );
-
-        setInputState(this, inputName, validationState);
-        return;
-      }
-
-      if (!previousInputState.isBlurred && inputState.isBlurred) {
```
marcdavi-es: In this case we set the `isValid` and `error` states to whatever they need to be.

src/components/collections/Form/component.spec.js
```diff
@@ -129,134 +123,6 @@ describe('<Form />', () => {
 
         expect(Object.entries).toHaveBeenCalledWith(wrapper.state());
       });
```
marcdavi-es: We need to test if `getFormsInputState` has been called for each entry in state, right?

src/components/collections/Form/utils/getFormInputsState.js
```diff
@@ -0,0 +1,56 @@
...
+import { setInputState } from './setInputState';
+
+/**
+ * @param {Object} input
+ * @param {string} inputName
+ * @param {Object} inputState
+ * @param {string} previousInputState
+ */
+export const getFormInputsState = (
+  input,
```
marcdavi-es: This is misleading, no?

This is `this` i.e. the form component.

src/components/collections/Form/utils/getFormInputsState.js
```diff
@@ -0,0 +1,56 @@
...
+
+  if (previousInputState.value !== inputState.value) {
+    const validationState = getValidationState(
+      input.props.validation[inputName],
+      inputState.value
+    );
+
+    setInputState(input, inputName, validationState);
+    return;
+  }
```
marcdavi-es: I'm not sure that this `if` block serves a purpose anymore?

src/components/collections/Form/utils/getFormInputsState.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const isRequiredState = getIsRequiredState(
+    input.props.validation[inputName],
+    inputState
+  );
+  const validationState = getValidationState(
+    input.props.validation[inputName],
+    inputState.value
+  );
```
marcdavi-es: `getIsRequiredState` and `getValidationState` can probably be merged into one thing, now that they are being called at the same time. They do roughly the same thing.

src/components/collections/Form/utils/getFormInputsState.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const errorState = getErroredState(validationState, isRequiredState);
+
+  setInputState(input, inputName, errorState);
+  return;
```
marcdavi-es: No need for this `return`. It's the same as having nothing at all.

src/components/collections/Form/utils/getValidationWithDefaults.js
```diff
@@ -10,7 +10,7 @@ import { DEFAULT_IS_REQUIRED_MESSAGE } from '../constants';
  */
 export const getValidationWithDefaults = ({
   getIsEmpty = value => !value && value !== 0,
-  getIsValid = Function.prototype,
+  getIsValid = () => true,
```
marcdavi-es: Do we want the default behaviour to validate anything?

Jonnygwi: I spoke to Julio about this one. He said that it makes sense for everything to be valid by default. There is a case where some inputs, in a form, have validation and others don't. After filling out the inputs there will be ticks next to valid inputs with validation and nothing next to inputs without validation although all of the inputs are valid values. Julio and I both agree that this could lead to confusion.

PR #616, opened 2019-04-29
TE-2165 Static google map

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2165)

Displays a static map until qualifying mouse events happen

Kapture 2019-04-29 at 13 20 43


src/components/elements/GoogleMap/component.js
```diff
@@ -1,33 +1,104 @@
...
+    this.img.addEventListener('mouseout', this.setIsMouseOverMapFalse);
+    this.setParentNodeWidth();
+  };
+
+  componentDidUpdate = (previousProps, previousState) =>
+    !this.state.isDynamicMapShowing &&
+    !previousState.isMouseOverMap &&
+    this.state.isMouseOverMap &&
+    global.setTimeout(
+      () => this.state.isMouseOverMap && this.setIsDynamicMapShowingTrue(),
```
Jonnygwi: I don't think you need this second `this.state.isMouseOverMap &&`.

marcdavi-es: Yeah we need it because the check runs half a second later. 

This makes sure we don't load the map if the mouse passes over the map momentarily - only if the user appears to be about to click.

Jonnygwi: Ahhh got it.. Nice! Approved!

PR #622, opened 2019-05-03
TE-1772 Fix slideshow loop and resize issues

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-1772)

* Solves image request loop.
* Sets breakpoints for the slideshow following the guidelines for the ticket.
* Removes load and resize handlers falling back on the breakpoints. 

### Notes
* By setting a fixed height for the component like described in the ticket, we assume that, if the image doesn't have enough height, a grey bar will be displayed at the bottom of the image in `IE 11`.

### Behaviour
Kapture 2019-05-03 at 11 14 42
Kapture 2019-05-03 at 11 17 52


src/components/media/Slideshow/component.js
```diff
@@ -8,40 +8,41 @@ import { Heading } from 'typography/Heading';
...
   componentDidMount = () => {
-    global.window.addEventListener('resize', this.handleResize);
+    this.handleImages(this.state.images);
   };
 
-  handleSlide = index => this.setState({ index });
-
-  handleResize = () => {
-    const imageHeight = getImageHeight(this.component);
+  handleImages = images => {
```
marcdavi-es: `handle*` is used for event handlers. There's no event being handled here.

If we need to adapt `props.images` and persist it in state, why not do

```js
  state = {
    adaptedImages: adaptImages(this.props.images),
```

?

src/components/media/Slideshow/utils/adaptImages.js
```diff
@@ -1,23 +1,13 @@
...
-        }
-  );
+export const adaptImages = images =>
+  images.map(({ alternativeText, url, sizes, srcSet, title }) => ({
+    original: url,
+    originalAlt: alternativeText,
+    originalTitle: title,
+    sizes,
+    srcSet,
+  }));
```
marcdavi-es: Is this just returning to the way this function was written before the last set of changes?

src/components/media/Slideshow/utils/setBrokenImagesToPlaceholders.js
```diff
@@ -0,0 +1,15 @@
+import { renderBlockPlaceholder } from './renderBlockPlaceholder';
+
+/**
+ * @param  {Object[]} adaptedImages
+ * @param  {string}   brokenImage
+ * @return {Object[]}
+ */
+export const setBrokenImagesToPlaceholders = (adaptedImages, brokenImage) =>
```
marcdavi-es: We use `set*` when we're setting something in state or another side effect. This should be something like `getAdaptedImagesWithPlaceholders`

src/components/media/Slideshow/utils/adaptImages.js
red-runner: @marcdavi-es Correct, since I'm not passing brokenImage this function can stay the way it used to be

PR #630, opened 2019-05-16
TE-2181 GoogleMaps: support an approximate location circle with a static map image

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2181)

Implements support for approximate locations on static map images.

Kapture 2019-05-16 at 19 35 33


red-runner:
> Looks like the static circle isn't aligned with the dynamic circle.
> 
> You can specify an `anchor` point in the query... https://developers.google.com/maps/documentation/maps-static/dev-guide#Markers

Agreed, I've aligned it accordingly. Attached new gif.

PR #632, opened 2019-05-20
TE-2169 Full Page menu subitems expand.

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2169)

When sub items are open in the full page menu the scroll moves depending on whether the sub item children are in the view.

### Any other notes
Kapture 2019-05-17 at 18 04 31


src/components/collections/Header/utils/HiddenMenu/component.js
```diff
@@ -0,0 +1,174 @@
...
+      activeAccordionItem.offsetParent.scrollTop = getScrollToHeight(
+        currentScrollHeight,
+        bottomOfActiveAccordionItem
+      );
+    }
+  };
+
+  createMenuRef = menu => (this.menu = menu);
+
+  onAccordionTitleClick = event => {
```
red-runner: Perhaps rename `onAccordionTitleClick` to `handleAccordionTitleClick` for consistency

src/components/collections/Header/utils/HiddenMenu/component.js
```diff
@@ -0,0 +1,174 @@
...
+        currentScrollHeight,
+        bottomOfActiveAccordionItem
+      );
+    }
+  };
+
+  createMenuRef = menu => (this.menu = menu);
+
+  handleAccordionTitleClick = event => {
+    if (this.state.activeAccordionItem) {
```
marcdavi-es: Which case does this cover?

Jonnygwi: After an accordion item has been clicked it remains in state so after multiple clicks to the same accordion item `componentDidUpdate` does not get fired. 

Accordion item clicked -> scroll location is updated -> user clicks the same item again (closes accordion) -> user scrolls -> clicks same item again -> no update to the scroll location.

PR #634, opened 2019-05-22
TE-2202 Refactor out lodash

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2202)

Remove Lodash as a dependency

### Any other notes
567 Bytes [debounce](https://bundlephobia.com/result?p=debounce@1.2.0) -> 1.9 Kilobytes [lodash.debounce](https://bundlephobia.com/result?p=lodash.debounce@4.0.8)
647 Bytes [Array.flat polyfill](https://bundlephobia.com/result?p=array-flat-polyfill@1.0.1) -> 1.4 Kilobytes [lodash.flatten](https://bundlephobia.com/result?p=lodash.flatten@4.4.0)
464 Bytes [uniqid](https://bundlephobia.com/result?p=uniqid@5.0.3) -> 856 Bytes [lodash.uniqueId](https://bundlephobia.com/result?p=lodash.uniqueid@4.0.1)
1.2 Kilobytes [get-value](https://bundlephobia.com/result?p=get-value@3.0.1) -> 4.4 Kilobytes [lodash.get](https://bundlephobia.com/result?p=lodash.get@4.4.2)
830 Bytes [fast-deep-equals](https://bundlephobia.com/result?p=fast-deep-equal@2.0.1) -> 10.1 Kilobytes [lodash.isequal](https://bundlephobia.com/result?p=lodash.isequal@4.5.0)

### DIFF
--- a/src/utils/is-nil/isNil.js
+++ b/src/utils/is-nil/isNil.js
@@ -1,0 +1,6 @@
+/**
+ * Is the value equal to undefined or null
+ * @param  {*} value
+ * @return {boolean}
+ */
+export const isNil = value => [undefined, null].includes(value);
--- a/src/utils/some/some.js
+++ b/src/utils/some/some.js
@@ -1,0 +1,20 @@
+import { getIsObjectEmpty } from './utils/getIsObjectEmpty';
+
+/**
+ * @param  {*}   collection
+ * @param  {Function} predicate
+ * @return {boolean}
+ */
+export const some = (collection, predicate) => {
+  const isObject = typeof collection === 'object' && collection !== null;
+
+  if (isObject && getIsObjectEmpty(collection)) {
+    return false;
+  }
+
+  if (isObject && predicate !== undefined) {
+    return Object.keys(collection).some(key => predicate(collection[key]));
+  }
+
+  return !!collection;
+};
--- a/src/utils/to-upper/toUpper.js
+++ b/src/utils/to-upper/toUpper.js
@@ -1,0 +1,6 @@
+/**
+ * @param  {string} string
+ * @return {string|undefined}
+ */
+export const toUpper = string =>
+  typeof string === 'string' ? string.toUpperCase() : undefined;
--- a/package.json
+++ b/package.json
@@ -119,7 +119,6 @@
     "is-html": "^1.1.0",
     "is-url": "^1.2.4",
     "libphonenumber-js": "^1.1.9",
-    "lodash": "^4.17.5",
     "react-dates": "^17.0.0",
     "react-google-maps": "^9.4.5",
     "react-image-gallery": "^0.8.7",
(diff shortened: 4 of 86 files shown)


src/utils/to-upper/toUpper.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * @param  {string} string
+ * @return {string|undefined}
+ */
+export const toUpper = string =>
+  typeof string === 'string' ? string.toUpperCase() : undefined;
```
red-runner: Perhaps returning an empty string would be better than `undefined` to avoid `toUpper(undefined).replace('awc', 'wcl')`

e.g
```
''.replace('s', 'a')  // ''
''.replace('', 'a') // a
undefined.replace('', 'a') // Uncaught TypeError
```

Jonnygwi: This isn't a legit use case for lodgify-ui as we are never calling a String.prototype method after the `toUpper` function is called.

src/utils/is-nil/isNil.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Is the value equal to undefined or null
+ * @param  {*} value
+ * @return {boolean}
+ */
+export const isNil = value => value == null;
```
marcdavi-es: Let's not use loose equals. Maybe...

```suggestion
export const isNil = value => [undefined, null].includes(value);
```

?

src/utils/some/some.js
```diff
@@ -0,0 +1,12 @@
...
+ * @param  {Function} [predicate=Function.prototype]
+ * @return {boolean}
+ */
+export const some = (collection, predicate = Function.prototype) => {
+  if (typeof collection === 'object' && collection !== null) {
+    return Object.keys(collection).some(key => predicate(collection[key]));
+  }
+
+  return false;
+};
```
marcdavi-es: I don't think this covers all the cases in `InputController`.

There we use `some` to return `true` if `value` is NOT one of `null`, `undefined`, `''`, `0`, `{}` or `{ startDate: null, endDate: null }`.

I think that is all of them.

src/components/inputs/InputController/utils/getIsDirtyValue.js
```diff
@@ -0,0 +1,15 @@
...
+  if (typeof value === 'object' && value !== null) {
+    const valueKeys = Object.keys(value);
+
+    return valueKeys.length > 0
+      ? !valueKeys.every(key => value[key] === null)
+      : false;
+  }
+
+  return !!value;
+};
```
marcdavi-es: Are we using this anywhere?

Jonnygwi: Yeye we are using it in `getIsDirtyValue.spec.js`

PR #642, opened 2019-05-31
TE-2253 Responsive image defaults to lazyload

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2253)

Makes all `ResponsiveImage` components to have lazyload by default. 

### Featured Property
featured_properties

### Featured RoomType
featured_roomtype

### Hero
hero

### Host Profile
host_profile

### Pictures
pictures

### Responsive Image
responsive_image

### Promotion 
promotion

### Thumbnail
thumbnail

### Homepage Hero
homepage_hero

### PropertyPageHero
propertypage_

### DIFF
--- a/src/components/layout/LazyLoader/component.js
+++ b/src/components/layout/LazyLoader/component.js
@@ -1,10 +1,12 @@
-import React, { PureComponent } from 'react';
+import React, { PureComponent, Fragment } from 'react';
 import { debounce } from 'debounce';
 import PropTypes from 'prop-types';
 
-import { getIsVisible } from './utils/getIsVisible';
+import { getIsTopComponentVisible } from './utils/getIsTopComponentVisible';
 import { getComponentToRender } from './utils/getComponentToRender';
 import { getComponentToRenderProps } from './utils/getComponentToRenderProps';
+import { getIsBottomComponentVisible } from './utils/getIsBottomComponentVisible';
+import { getComponentPosition } from './utils/getComponentPosition';
 
 /**
  * A wrapper which intercepts props being passed to a child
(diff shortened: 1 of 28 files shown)


src/components/layout/LazyLoader/component.js
```diff
@@ -88,7 +88,11 @@ export class Component extends PureComponent {
       lazyProps
     );
 
-    return <div ref={this.createComponentRef}>{componentToRender}</div>;
+    return (
+      <div className="responsive-image-wrapper" ref={this.createComponentRef}>
```
marcdavi-es: We don't want to add a class related to `ResponsiveImage` to `LazyLoader`. `LazyLoader` is a general purpose component.

red-runner:
@marcdavi-es This now works without adding styles 🎉

src/components/layout/LazyLoader/component.js
```diff
@@ -88,7 +88,13 @@ export class Component extends PureComponent {
       lazyProps
     );
 
-    return <div ref={this.createComponentRef}>{componentToRender}</div>;
+    return (
+      <Fragment>
+        <div ref={this.createComponentRef} />
+        {componentToRender}
+        <div ref={this.createComponentRef} />
```
Jonnygwi: I don't think you can set a reference to two components like this. The bottom of this component will be that of the first div.. Not the last. I think we need to make two refs and update the `getShouldComponentRender` method.

red-runner:
> One thing inline. Also is the thumbnail gif up to date as it looks like the image isn't centered as it was before the change.

@Jonnygwi You were right about the thumbnail; with the new implementation doesn't happen. Uploaded new gif.

PR #673, opened 2019-07-04
TE-2393 "Lazy" images are being loaded if component offset is zero

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2393)

Prevents `LazyLoader` to render responsive images if component is on position 0.

### Before
before

### After
after

### DIFF
--- a/src/components/layout/LazyLoader/utils/getIsTopComponentVisible.spec.js
+++ b/src/components/layout/LazyLoader/utils/getIsTopComponentVisible.spec.js
@@ -22,4 +22,12 @@ describe('getIsVisible', () => {
       expect(actual).toBe(false);
     });
   });
+
+  describe('when top offset is zero', () => {
+    it('should return false', () => {
+      const actual = getIsTopComponentVisible({ top: 0 }, 1000);
+
+      expect(actual).toBe(false);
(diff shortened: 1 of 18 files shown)


src/components/layout/LazyLoader/utils/getIsTopComponentVisible.spec.js
```diff
@@ -22,4 +22,12 @@ describe('getIsVisible', () => {
       expect(actual).toBe(false);
     });
   });
+
+  describe('when bottom offset is zero', () => {
```
Jonnygwi: `'when bottom offset is zero'` -> `'when top offset is zero'`

src/components/media/ResponsiveImage/component.spec.js
```diff
@@ -39,6 +39,7 @@ describe('<ResponsiveImage />', () => {
     it('should call the Image constructor', () => {
       const wrapper = getWrappedResponsiveImage({
         placeholderImageUrl: 'ayyy',
+        isLazyLoaded: false,
```
Jonnygwi: Why are we adding this here?

PR #676, opened 2019-07-09
TE-2419 Isolate new `Label` component and extend.

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2419)

 * Isolates a new `Label` component. 

 * `PropertySearchResult` and `PropertySearchResultList` consume `Label`.

 * `PropertySearchResult`, `PropertySearchResultList` and `Label` have `pricePerPeriodPreposition` as a new prop: which will allow us to translate `from` preposition.

### Exposed cases for Label 

Screenshot 2019-07-10 at 16 05 23
Screenshot 2019-07-10 at 16 05 27
Screenshot 2019-07-09 at 18 17 31
Screenshot 2019-07-09 at 18 17 41
Screenshot 2019-07-09 at 18 18 24
Screenshot 2019-07-10 at 16 03 41
Screenshot 2019-07-10 at 16 03 50

### Hovering 

hoverings

### Proof `PropertySearchResult` woks AE.

Screenshot 2019-07-09 at 18 20 04

### Proof semantic-ui pointing doesn't fit

Screenshot 2019-07-10 at 11 31 56


src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+import { FlexContainer } from 'layout/FlexContainer';
+import { Paragraph } from 'typography/Paragraph';
+
+/**
+ * The standard widget to display a price Label.
+ */
+// eslint-disable-next-line jsdoc/require-jsdoc
+export const Component = ({
+  isMapMarker,
+  isTriggeredOnHover,
```
Jonnygwi: Could this prop not just be called `isHovered`?

Jonnygwi: This change would removing all references to "triggered on"

src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+  pricePerPeriod: '',
+  pricePerPeriodPreposition: '',
+};
+
+Component.propTypes = {
+  /** Is the Label displayed as a marker. */
+  isMapMarker: PropTypes.bool,
+  /** Is the Label getting triggered by hovering another component. */
+  isTriggeredOnHover: PropTypes.bool,
+  /** The function to call when the link is clicked.
```
Jonnygwi: ` /** The function to call when the link is clicked.` -> `/** The function to call when the label is clicked.`

src/components/elements/Label/component.js
red-runner: In this case, I got the reference from `Submenu`, I think triggered makes sense since it gets the hover effect from an external component and not from the component itself. `isHovered` makes me think more about hovering the `Label`. Thoughts?

src/components/elements/Label/Readme.md
```diff
@@ -0,0 +1,71 @@
+```jsx
+<Label 
+  pricePerPeriod="€200" 
+  periodText="per month" 
+  pricePerPeriodPreposition="from" 
+  onClick={() => console.log('plo')}
+/>
+```
+
+### Wihtout period text
```
marcdavi-es:
```suggestion
### Without period text
```

src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+import { Label } from 'semantic-ui-react';
+import getClassNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import { Heading } from 'typography/Heading';
+import { FlexContainer } from 'layout/FlexContainer';
+import { Paragraph } from 'typography/Paragraph';
+
+/**
+ * The standard widget to display a price Label.
```
marcdavi-es: I think this format is what we use for the bigger composed widgets. For elements we normally do something like

```
* A price label displays pricing information relating to nearby content.
```

or something?

src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+          <Heading size="medium">{pricePerPeriod}</Heading>
+        </FlexContainer>
+        <Label> {`${periodText}`}</Label>
+        {isMapMarker && <div className="marker-arrow" />}
+      </Fragment>
+    )}
+  </Label>
+);
+
+Component.displayName = 'Label';
```
marcdavi-es: Since the props are closely tied to displaying pricing information only, I guess we should call the component `PriceLabel` ?

Alternatively we could make it more general purpose such that it accepts any children. Then `Label` is appropriate.

src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+import { Heading } from 'typography/Heading';
+import { FlexContainer } from 'layout/FlexContainer';
+import { Paragraph } from 'typography/Paragraph';
+
+/**
+ * The standard widget to display a price Label.
+ */
+// eslint-disable-next-line jsdoc/require-jsdoc
+export const Component = ({
+  isMapMarker,
```
marcdavi-es: I wouldn't tie this to being a map marker. It is true that we will use it as a map marker in this instance, but really what we want is the following variations:

- `isPointing`: which gives it the little pointer at the bottom
- `hasShadow`: which gives it the shadow
- `isActive`: which gives it the filled background

This makes is more general purpose and composable.

src/components/elements/Label/component.js
marcdavi-es: I would use `isActive` and control hover state in the same way I do for `Marker`. See [this guy](https://github.com/lodgify/lodgify-ui/pull/677/commits/e2b4aeb669670a2c3286ed3774f046798786534c#diff-5e8404527082a570688b6a16c5d3d10bR11) in my PR

src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+  isTriggeredOnHover,
+  onClick,
+  periodText,
+  pricePerPeriod,
+  pricePerPeriodPreposition,
```
marcdavi-es: I would use `pricePerPeriodPrefix`. 'prefix' is pretty commonly used in this context.

src/components/elements/Label/component.js
```diff
@@ -0,0 +1,74 @@
...
+}) => (
+  <Label
+    className={getClassNames('search-result-label', {
+      'is-map-marker': isMapMarker,
+      'has-period-text': periodText,
+      'is-triggered-on-hover': isTriggeredOnHover,
+    })}
```
marcdavi-es: Rather than rely heavily on extra styles, let's use the semantic-ui-react component API more extensively. For example, the `pointing` prop is something we could use here.

src/components/elements/PriceLabel/component.js
```diff
@@ -0,0 +1,117 @@
...
+
+    return (
+      <div
+        className={getClassNames('ui', 'label', 'price-label', {
+          'is-active': isActive || this.state.isActive,
+          'is-pointing': isPointing,
+          'has-period-text': periodText,
+          'has-shadow': hasShadow,
+        })}
+        onClick={onClick}
```
marcdavi-es: I think `props.onClick` ought to be passed `props.name` as a parameter like `onChange`. So we can easily identify which `PriceLabel` has been clicked.

src/components/elements/PriceLabel/component.js
```diff
@@ -0,0 +1,117 @@
...
+  name: undefined,
+  onChange: Function.prototype,
+  onClick: Function.prototype,
+  periodText: '',
+  pricePerPeriod: '',
+  pricePerPeriodPrefix: '',
+};
+
+Component.propTypes = {
+  /** Does the Label have a shadow. */
```
marcdavi-es: Have a check for lingering `Label`s. Needs to be `PriceLabel`

src/components/elements/PriceLabel/component.js
```diff
@@ -0,0 +1,117 @@
...
+  onChange: PropTypes.func,
+  /** The function to call when the Label is clicked.
+   *  @param {Object} event
+   */
+  onClick: PropTypes.func,
+  /** The text describing the pricing period. */
+  periodText: PropTypes.string,
+  /** The price per period of the property, with currency symbol. */
+  pricePerPeriod: PropTypes.string,
+  /** The text preposition to display along with the pricePerPeriod. */
```
marcdavi-es: `preposition` -> `prefix`

src/components/elements/PriceLabel/Readme.md
```diff
@@ -0,0 +1,114 @@
...
+
+```jsx
+<div>
+  <PriceLabel hasShadow isPointing pricePerPeriod="€200"  onClick={() => console.log('plo')}/>
+</div>
+```
+
+### Usage
+
+#### Activated by external hover
```
marcdavi-es: No need for this very specific example. Just have

```
### States

#### Active
```

Like in `Marker`

src/components/elements/PriceLabel/Readme.md
```diff
@@ -0,0 +1,114 @@
...
+    isPointing 
+    pricePerPeriod="€200"
+    pricePerPeriodPrefix="from"  
+    periodText="per month" 
+    onClick={() => console.log('plo')}
+  />
+</div>
+```
+
+#### Pointing without period text
```
marcdavi-es: No need for this example.

src/components/elements/PriceLabel/Readme.md
```diff
@@ -0,0 +1,114 @@
...
+    isPointing 
+    pricePerPeriod="€200"
+    pricePerPeriodPrefix="from"  
+    periodText="per month" 
+    onClick={() => console.log('plo')}
+  />
+</div>
+```
+
+#### Has shadow without period text
```
marcdavi-es: No need for this example.

src/components/elements/PriceLabel/Readme.md
```diff
@@ -0,0 +1,114 @@
...
+  position: 'relative',}}
+>
+  <PriceLabel isPointing pricePerPeriod="€200"  onClick={() => console.log('plo')}/>
+</div>
+```
+
+#### Has shadow
+
+```jsx
+<div>
```
marcdavi-es: Remove this div.

src/components/elements/PriceLabel/Readme.md
```diff
@@ -0,0 +1,114 @@
+```jsx
+<PriceLabel 
+  pricePerPeriod="€200" 
+  periodText="per month" 
+  pricePerPeriodPrefix="from" 
+  onClick={() => console.log('plo')}
```
marcdavi-es: In all of these cases, I would just pass `console.log` as `onClick`. Then the user can see in the console what `onClick` receives.

`'plo'` doesn't tell us anything.

`onClick={console.log}`

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,111 @@
     font-weight: @basicFontWeight;
   }
 
+  /* Price Label */
+
+  &.price-label {
+    background-color: var(@themeActionColorIdentifier, @themeActionColorDefault);
```
marcdavi-es: This is missing the fallback for browsers which don't support `var`

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,111 @@
...
+    border-radius: @priceLabelBorderRadius;
+    padding: @priceLabelPadding;
+    padding-top: @priceLabelPaddingTop;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    position: relative;
+    float: left;
```
marcdavi-es: Why are we using `float`?

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,111 @@
...
+          top: @pointerArrowTopOverflow;
+        }
+      }
+
+      &.has-period-text .pointer-arrow {
+        top: @pointerArrowContainerWithoutPeriodTextTopOffset;
+      }
+
+      &.is-active {
+        background-color: var(@themeActionColorIdentifier, @themeActionColorDefault) !important;
```
marcdavi-es: Needs the fallback.

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,111 @@
...
+        transform: translateY(@pointerHoverTranslationY);
+        transition: @priceLabelTransitionOnHover;
+
+        .pointer-arrow:after {
+          background-color: var(@themeActionColorIdentifier, @themeActionColorDefault) !important;
```
marcdavi-es: Needs the fallback.

src/styles/semantic/themes/livingstone/elements/label.overrides
red-runner: For 2. The initial state for label is coloured yellow. See https://youtrack.lodgify.net/agiles/86-18/87-1266?issue=TE-2419 

Then for the map marker works the other way around. I understood as the `pointing` version being a modification of the original.

src/components/elements/PriceLabel/component.js
```diff
@@ -0,0 +1,122 @@
...
+              <Heading size="medium">{pricePerPeriod}</Heading>
+              {isPointing && <div className="pointer-arrow" />}
+            </Fragment>
+          ) : (
+            <Fragment>
+              <FlexContainer alignItems="center" flexDirection="row">
+                <Paragraph>{pricePerPeriodPrefix}</Paragraph>
+                <Heading size="medium">{pricePerPeriod}</Heading>
+              </FlexContainer>
+              <Label> {`${periodText}`}</Label>
```
marcdavi-es: This is the same as 

```jsx
<Label>{periodText}</Label>
```

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,113 @@
     font-weight: @basicFontWeight;
   }
 
+  /* Price Label */
+
+  &.price-label {
+    background-color: @primaryColor;
```
marcdavi-es: This should be `@themeActionColorDefault`

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,113 @@
...
+          top: @pointerArrowTopOverflow;
+        }
+      }
+
+      &.has-period-text .pointer-arrow {
+        top: @pointerArrowContainerWithoutPeriodTextTopOffset;
+      }
+
+      &.active {
+        background-color: @primaryColor;
```
marcdavi-es: This should be `@themeActionColorDefault`

src/styles/semantic/themes/livingstone/elements/label.overrides
```diff
@@ -76,6 +76,113 @@
...
+        background-color: var(@themeActionColorIdentifier, @themeActionColorDefault) !important;
+        transform: translateY(@pointerHoverTranslationY);
+        transition: @priceLabelTransitionOnHover;
+
+        .pointer-arrow:after {
+          background-color: @primaryColor;
```
marcdavi-es: This should be `@themeActionColorDefault`

PR #677, opened 2019-07-09
TE-2420 marker component

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2420)

Creates a new `Marker` component

marker

### DIFF
--- a/src/components/elements/Marker/component.js
+++ b/src/components/elements/Marker/component.js
@@ -1,0 +1,58 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import getClassNames from 'classnames';
+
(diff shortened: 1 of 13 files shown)


src/components/elements/Marker/component.js
```diff
@@ -0,0 +1,58 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import getClassNames from 'classnames';
+
+import { getToggledState } from './utils/getToggledState';
+
+/**
+ * A marker marks a point on a plane.
```
red-runner: Maybe 

`A marker which marks a point on a plane.`
or
`A marker that marks a point on a plane.`

PR #680, opened 2019-07-10
TE-2425 Add shadow on hover to `PropertySearchResult`

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2425)

Adds shadow on hover to `PropertySearchResult` though active state

### Any other notes
Kapture 2019-07-10 at 20 34 30


src/components/general-widgets/PropertySearchResult/component.js
```diff
@@ -17,82 +19,109 @@ import { getPropertyAmenities } from './utils/getPropertyAmenities';
...
+      propertyType,
+      propertyUrl,
+      propertyUrlTarget,
+      ratingNumber,
+    } = this.props;
+
+    return (
+      <div
+        className={getClassNames('ui', 'card', 'has-search-result', {
+          active: isActive || this.state.isActive,
```
Jonnygwi: I think our class names normally follow the property name for example this would become 'is-active'

marcdavi-es: Mmm I see what you mean but I think there are two possibilities. In many places we following `semantic-ui` conventions for things like `active`.

I have used `active` for `Marker` in a PR this week. I'm ok with this.

src/styles/semantic/themes/livingstone/views/card.overrides
```diff
@@ -113,81 +113,92 @@
...
-    height: @searchResultHeight;
     margin: 0 @searchResultMargin @searchResultMargin 0;
+    height: @searchResultHeight;
     width: @searchResultWidth;
+    transition: @searchResultBoxShadowTransition;
 
-    @media @tinyScreen {
-      width: 100%;
-      margin-right: 0;
+    &.active {
```
Jonnygwi: 'is-active' note

src/styles/semantic/themes/livingstone/views/card.overrides
```diff
@@ -113,81 +113,92 @@
...
-          padding: @searchResultLabelPeriodTextPadding;
-          background: transparent;
-          border-radius: @searchResultLabelPeriodBorderRadius;
-          box-shadow: @searchResultLabelPeriodBoxShadow;
-          color: @black;
+      > .flex-container {
+        position: relative;
+
+        > .ui.raised.segment {
+          background-color: var(@themeActionColorIdentifier, @themeActionColorDefault);
```
marcdavi-es: This guy is missing the fallback.

src/components/general-widgets/PropertySearchResult/component.js
Jonnygwi: We should use one or the other no?

red-runner: I followed the pattern used in `Marker` to avoid confusion.

src/utils/get-toggled-state/getToggledState.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * @param  {Object}  previousState
+ * @param  {boolean} previousState.isActive
+ * @return {Object}
+ */
+export const getToggledState = ({ isActive }) => ({ isActive: !isActive });
```
marcdavi-es: Nice idea. 
1. Now that it's global lets give it a more specific name, since it isn't a general purpose state toggler. Maybe `getToggledIsActiveState`
2. Let's consume this guy in `Marker` and `PriceLabel`

src/components/general-widgets/PropertySearchResult/component.js
marcdavi-es: We use `.active` in a bunch of places and `.is-active` doesn't exist yet.

Screenshot 2019-07-11 at 15 06 15

Jonnygwi: Okay. So with class naming we are keeping semantic ui's naming convention for states.

PR #682, opened 2019-07-12
TE-2428 PropertySearchResult exposes an `isCompact` prop

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2428)

Exposes `isCompact` prop for `PropertySearchResult`

### Behaviour

Kapture 2019-07-11 at 21 35 18

Screenshot 2019-07-12 at 09 18 45

### DIFF
--- a/src/components/general-widgets/PropertySearchResult/component.js
+++ b/src/components/general-widgets/PropertySearchResult/component.js
@@ -12,7 +12,6 @@ import { Heading } from 'typography/Heading';
 import { FlexContainer } from 'layout/FlexContainer';
 import { PriceLabel } from 'elements/PriceLabel';
 
-// import { getPriceLabelMarkup } from './utils/getPriceLabelMarkup';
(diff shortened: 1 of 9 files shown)


src/components/general-widgets/PropertySearchResult/component.js
```diff
@@ -180,6 +194,8 @@ Component.propTypes = {
   imageUrl: PropTypes.string.isRequired,
   /** Is the property search result in an active state. */
   isActive: PropTypes.bool,
+  /** Is the property search result showing in compact display. */
```
Jonnygwi: Maybe something like this is more of a description? `/** Is the property search result only taking up as much space as necessary. */`

Jonnygwi: `/** Is the property search result displayed with a compact layout. */`

PR #694, opened 2019-07-23
TE-2473 Extend `PhoneInput`

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2473)

Extends `PhoneInput` with a dropdown for country selection

phone-input


Jonnygwi:
Looks good. Just a couple questions.

- Does `src/components/inputs/PhoneInput/utils/CountrySelectComponent/component.js` need to export a `PureComponent`? It could just be a pure function no?

- Can we mock the phone input component to avoid having such a massive snapshot?

marcdavi-es:
@Jonnygwi Thanks

- It has to be a class because `react-phone-number-input` creates refs on it behind the scenes. Functional components can't get refs.
- The snapshot isn't _that_ big and it contains some quite specific stuff because we compose the overall thing from a bunch of our own components. I think it's good to have that committed so we can pick up on potential breaking changes to e.g. `Dropdown`

Jonnygwi:
@marcdavi-es great. Thanks for answering my questions. In that case I'll give it an approve and merge it in. 👍

PR #707, opened 2019-08-07
TE-2502 `GoogleMap` add flag for drag events

marcdavi-es:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2502)

Passes a boolean drag event flag to `GoogleMap.props.onBoundsChange` 

### Any other notes
An update in `react-phone-number-input` meant I had to update some test cases.


src/utils/google-map-react/components/GoogleMapReact/utils/getCenterFromBounds.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * @param  {Object} bounds
```
Jonnygwi:
```
   * @param {Object}  bounds
   * @param {number}  bounds.east
   * @param {number}  bounds.north
   * @param {number}  bounds.south
   * @param {number}  bounds.west
```

Shouldn't we document this function like the above?

PR #721, opened 2019-09-18
Te 2628 location reset

andream-lodgify:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2628)

Allow the dropdown to clear its value after choose one. 

### Any other notes
In order to achieve the task was needed to update semantic-ui-react.This triggered the refresh of the snapshot and the refactor of the sidebar test

Dropwdown

### DIFF
--- a/src/components/inputs/Dropdown/component.js
+++ b/src/components/inputs/Dropdown/component.js
@@ -16,6 +16,7 @@ import { getIsOpenAfterChange } from './utils/getIsOpenAfterChange';
 import { adaptOptions } from './utils/adaptOptions';
 import { getHasImages } from './utils/getHasImages';
 import { isValueValid } from './utils/isValueValid';
+import { getIcon } from './utils/getIcon';
 
 /**
  * A dropdown allows a user to select a value from a series of options.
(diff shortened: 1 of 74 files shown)


src/components/elements/Submenu/component.js
```diff
@@ -29,7 +29,7 @@ export const Component = ({
 }) => (
   <Dropdown
     className={isTriggerUnderlined ? 'underlined' : ''}
-    icon={<Icon name={ICON_NAMES.CARET_DOWN} size="small" />}
+    icon={<Icon className="icon" name={ICON_NAMES.CARET_DOWN} size="small" />}
```
Jonnygwi: Are you sure that this `Icon` needs to have that `className`?
In the component of `<Icon />` I can see that it's being added by the `getClassNames` function.

src/components/inputs/Dropdown/component.js
```diff
@@ -192,6 +180,8 @@ Component.propTypes = {
     PropTypes.number,
     PropTypes.string,
   ]),
+  /** A dropdown is clearable  */
```
Jonnygwi: I think this is more explicit. `/** The dropdown's value can be cleared.  */`. What do you think?

src/components/inputs/Dropdown/component.spec.js
```diff
@@ -177,7 +162,17 @@ describe('<Dropdown />', () => {
       });
     });
 
-    describe('if `isDisabled` prop is passed', () => {
+    describe('if `isClearable` prop is passed false', () => {
```
Jonnygwi: `'if `isClearable` prop is false'`

src/components/inputs/Dropdown/utils/getIcon.js
```diff
@@ -0,0 +1,6 @@
+import { ICON_NAMES } from 'elements/Icon';
+
+import { isValueValid } from './isValueValid';
+
+export const getIcon = (value, isClearable = true) =>
```
Jonnygwi: We normally add JSDocs to out utils.

```
/**
 * @param  {*}  value
 * @param  {*} previousState.isActive
 * @return {string}
 */
```

PR #728, opened 2019-09-30
TE-2669 Fix missing translations

red-runner:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2669)

Implements a prop for `More info` translation for both `RoomType` and `RoomTypes` component.
Implements a prop for the conjunction which unites the last and the penultimate amenity items

### Any other notes
Amenities
More_info

### DIFF
--- a/src/components/property-page-widgets/Amenities/Readme.md
+++ b/src/components/property-page-widgets/Amenities/Readme.md
@@ -121,7 +121,7 @@ const amenities = [
   },
 ];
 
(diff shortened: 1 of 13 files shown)


red-runner:
> Code looks good but I can see that it is saying `undefined` before the last item on the miscellaneous group.

@Jonnygwi You were right, something was missing. Attaching proof of fixing: 

Screenshot 2019-09-30 at 15 31 15

PR #730, opened 2019-09-30
TE-2659: Remove label behvaiour from InputController and Dropdown

andream-lodgify:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2659)

Remove label behvaiour from InputController and Dropdown


styleguide/styleguide-components/LogoRenderer/component.js
```diff
@@ -1,15 +1,20 @@
...
-  <a href={getIsDevelopmentServer() ? '/' : '/lodgify-ui/'}>
-    <img
-      alt="Lodgify UI"
-      src={logo.default}
-      style={{ display: 'block', margin: `0${hasNoMargin ? '' : ' auto'}`, width: '110px', }}
-    />
-  </a>
-);
+export const component = ({ hasNoMargin }) =>
+  console.log(logo) || (
```
Jonnygwi: Let's delete this. 

I've also made this change in my PR so you can drop this one if it goes in first.

PR #754, opened 2019-10-21
TE-2746: input refactor

huvber:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2746)
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2781)

Now most of the input component are controlled by default and doesn't have internal state anymore. This is a kind of breaking change.


src/components/collections/CheckboxInputSegment/Readme.md
```diff
@@ -1,8 +1,40 @@
...
+
+```jsx
+const { checkboxes } = require('./mock-data/checkboxes');
+const Controller = () => {
+  const [values, setValues] = React.useState({});
+  const onChange = ({ name, value }) => {
+    setValues({ ...values, [name]: value });
+  };
+  const controlledCheckboxes = checkboxes.map(c => ({
+    ...c,
```
Jonnygwi: Let's not use `c` here. rename it to `checkbox` for easier reading.

src/components/collections/CheckboxInputSegment/Readme.md
```diff
@@ -1,8 +1,40 @@
...
+  return (
+    <React.Fragment>
+      <CheckboxInputSegment
+        heading="Amenities"
+        checkboxes={controlledCheckboxes}
+      />
+      <p>Selected Amenities:</p>
+      <ul>
+        {Object.keys(values)
+          .filter(k => values[k])
```
Jonnygwi: Same for here.

src/components/general-widgets/SearchBar/component.js
```diff
@@ -22,12 +21,22 @@ import { getSearchBarModal } from './utils/getSearchBarModal';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export class Component extends PureComponent {
-  state = {
-    dates: getInitialValue(this.props.datesInputValue),
-    guests: getInitialValue(this.props.guestsInputValue),
-    location: getInitialValue(this.props.locationInputValue),
-    willLocationDropdownOpenAbove: this.props.willLocationDropdownOpenAbove,
-  };
+  constructor(props) {
```
Jonnygwi: Why does this have to be in a constructor?

src/components/general-widgets/SearchBar/utils/getFormFieldMarkup.js
```diff
@@ -9,15 +9,16 @@ import { DateRangePicker } from 'inputs/DateRangePicker';
...
  * @param  {Function} props.datesInputOnFocusChange
  * @param  {Function} props.getIsDayBlocked
- * @param  {string}   props.guestsInputValue
- * @param  {string}   props.locationInputValue
  * @param  {Object[]} props.locationOptions
  * @param  {Node}     props.searchButton
+ * @param  {Object} state
+ * @param  {string} state.location
+ * @param  {Object} state.dates
+ * @param  {number} state.guests
```
Jonnygwi: The three above state properties should have the same naming as the props that you've just removed for consistency

src/components/inputs/Checkbox/Readme.md
```diff
@@ -42,15 +41,33 @@
...
+
+### Controlled checkbox
+
+```jsx
+const Controller = () => {
+  const [value, setValue] = React.useState(false);
+  return (
+    <Checkbox
+      isChecked={value}
+      onChange={(name, _value) => {
```
Jonnygwi: `_value`?

src/components/inputs/InputController/component.js
```diff
@@ -1,154 +1,114 @@
...
-  inputOnChangeFunctionName: 'onChange',
-  isCompact: false,
-  isFluid: false,
-  isFocused: false,
-  label: null,
-  mapValueToProps: value => ({ value }),
-  value: undefined,
+  return (
+    <Input
+      className={cx({
```
Jonnygwi: `cx` -> `getClassNames` please. cx doesn't mean anything

src/components/general-widgets/SearchBar/component.js
huvber: I think we already have this discussion before. I like to add the constructor in general because is javascript. state (or whatever property) defined in the body of the class is something that babel configuration allow you to do. Anyway I put it in this way because otherwise I had some problem with some tests

src/components/general-widgets/SearchBar/utils/getFormFieldMarkup.js
huvber: I don't understand why. If I do like you ask I have to do the same modification in the parent component state and doesn't have so much sense because they are not props, they are just state. Actually I don't think I like the fact that for this things there are function and not components.

src/components/inputs/Checkbox/Readme.md
huvber: Is to avoid collisions to the state one

src/components/inputs/InputController/component.js
huvber: neither `getClassNames`. Is something that in this project was renaming for reason that I don't understand. If you look the documentation of the library you will see that they suggest to call it like that
https://github.com/JedWatson/classnames#alternate-bind-version-for-css-modules

Actually in my opinion is even worst to rename some library because make it more difficult to recognize and to work with. 

__Why you didn't call `moment` `getMoment`? for example__

src/components/inputs/Checkbox/Readme.md
Jonnygwi: rename the `isChecked={value}` to `isChecked={isCheckedValue}` if that's easier and then rename `_value` to `value`.

src/components/inputs/InputController/component.js
Jonnygwi: Let's rename it to `cx` in the whole project in that case

src/components/general-widgets/SearchBar/component.js
Jonnygwi: I'm happy with this being in a constructor as it will eventually be refactored out when we use hooks.

PR #756, opened 2019-10-23
TE-2742 Counter dropdown

Jonnygwi:
[Related YouTrack issue](https://youtrack.lodgify.net/issue/TE-2742)

Expose CounterDropdown to the api

### Any other notes
Kapture 2019-10-23 at 13 15 46


src/components/inputs/CounterDropdown/component.js
```diff
@@ -0,0 +1,85 @@
...
+        clearable
+        icon={<Icon name={ICON_NAMES.CARET_DOWN} />}
+        onBlur={() => setIsOpen(false)}
+        onFocus={() => setIsOpen(true)}
+        open={isOpen}
+        options={[
+          <FlexContainer flexDirection="row" justifyContent="space-between">
+            <Paragraph>{counterLabel}</Paragraph>
+            <Counter
+              initialValue={initialCounterValue}
```
huvber: Since our inputs will be controlled I think we don't need `initialValue` we should just use value. Otherwise can bring update errors.

PR #775, opened 2019-11-13
WEB-497: Increase Header text heading font

huvber:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-497)

Increase the size of the title in the header when it's text.

### Any other notes
I also refactor the Logo inside the Header converting that in a component from a function

|before|after|
|---|---|
|image|image|


src/styles/semantic/themes/livingstone/elements/header.overrides
```diff
@@ -29,6 +29,10 @@
 
 /* Sizing */
 
+.ui.header.header-title {
+  font-size: @headingTtitleFontSize;
```
Jonnygwi:
```suggestion
  font-size: @headingTitleFontSize;
```

src/styles/semantic/themes/livingstone/elements/header.variables
```diff
@@ -33,6 +33,8 @@
 
 /* Sizing */
 
+@headingTtitleFontSize: 1.43rem;
```
Jonnygwi:
```suggestion
@headingTitleFontSize: 1.43rem;
```

PR #778, opened 2019-11-13
WEB-651 Hide the main icons guests/bedrooms/bathrooms when property has room types.

red-runner:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-651)

Exposes a new prop to control whether the `propertyMainCharacteristics` are shown or not, removing the markup accordingly.

Kapture 2019-11-13 at 15 49 00

### DIFF
--- a/src/components/property-page-widgets/Description/component.js
+++ b/src/components/property-page-widgets/Description/component.js
@@ -22,6 +22,7 @@ import { getDescriptionTextMarkup } from './utils/getDescriptionTextMarkup';
  */
 // eslint-disable-next-line jsdoc/require-jsdoc
 export const Component = ({
+  arePropertyMainCharacteristicsShown,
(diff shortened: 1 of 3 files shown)


src/components/property-page-widgets/Description/component.js
```diff
@@ -35,36 +36,38 @@ export const Component = ({
...
+                key={buildKeyFromStrings(text, index)}
+                mobile={6}
+              >
+                <Icon labelText={text} name={iconName} />
+              </GridColumn>
+            )
+          )}
+        </ShowOn>
+      </GridColumn>
+    )}
```
huvber: shouldn't be better to have a `Characteristics` component  that just take the array of characteristics and do the magic instead of having all this markup here?

PR #783, opened 2019-11-18
Web 155 description show more

huvber:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-155)

Add a 'show more' button under the text of the description that allow to enlarge that once. 

### Any other notes

**In this PR you will find two important things I would like we focus on:**
1. you will find a test for the `MultiParagraph` component that has 100% coverage but test only behaviour and doesn't use snapshot but instead **data-testid**. This test was written before even implement the component. It should be an example of how to write tests without checking any implementation ad all (snapshot testing which tag are you use so they actually check implementation). 

2. You will find in the `utils` folder a `testid` folder. This folder expose two twins function that allow you to easily implement components with **data-testid** and test them.

description


src/components/typography/MultiParagraph/component.js
```diff
@@ -0,0 +1,56 @@
...
+  );
+};
+
+Component.displayName = 'MultiParagraph';
+Component.defaultProps = {
+  showMoreLabel: SHOW_MORE,
+  isShowingAll: false,
+};
+Component.propTypes = {
+  /** text to show or hide */
```
red-runner: Here, look at other examples of propTypes descriptions and try to match them. Also (Start with a capital  and end with a dot.)

huvber: done

PR #785, opened 2019-11-22
WEB-682 Expose Lightbox component

red-runner:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-682)

* Exposes a brand new `LightBox` component. ✨ 
* Removes `isDark` from `Modal` and exposes `className` instead; since there is 1 use case for `isDark` and dark background needs to be applied to `Modal Content` instead of `Modal Dimmer`.
* Introduces slight style modifications from user story, the aim of which is to refine the appearance of the ui when dark background are used.

* Introduces `--coverage` flag to `test:watch`; this triggering instanbul coverage report `exclusively` for the files that changed every time.

### Behaviour
Kapture 2019-11-22 at 11 09 45

6b5f80da4f564b4f895d03068854dc49

Screenshot 2019-11-22 at 11 34 34

### DIFF
--- a/src/components/media/LightBox/component.js
+++ b/src/components/media/LightBox/component.js
@@ -1,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Modal } from 'elements/Modal';
+import { Slideshow } from 'media/Slideshow';
+import { testidFactory } from 'utils/testid';
+
+const TEST_ID_PREFIX = 'lightbox';
+
+const testid = testidFactory(TEST_ID_PREFIX);
+
+/**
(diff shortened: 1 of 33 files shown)


src/components/media/LightBox/component.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+Component.displayName = 'LightBox';
+
+Component.defaultProps = {
+  trigger: null,
+};
+
+Component.propTypes = {
+  slideShowImages: PropTypes.arrayOf(
```
Jonnygwi: We need a description on this prop

PR #794, opened 2019-11-27
WEB-695 Add default background

Jonnygwi:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-695)

MinimalPromotion has a background colour so that when there is no image prop the text is still visible.

### Any other notes
image

### DIFF
--- a/src/components/property-page-widgets/RoomType/component.js
+++ b/src/components/property-page-widgets/RoomType/component.js
@@ -224,26 +224,6 @@ Component.propTypes = {
     })
   ).isRequired,
   /** Is the component showing placeholders to reserve space for content which will appear. */
-  isShowingPlaceholder: PropTypes.bool,
-  /**
-   * Is the user on a mobile device.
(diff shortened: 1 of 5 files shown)


src/components/property-page-widgets/RoomType/component.js
```diff
@@ -256,6 +236,26 @@ Component.propTypes = {
...
     })
   ).isRequired,
+  /**
+   * Is the user on a mobile device.
+   * Provided by `withResponsive` so ignored in the styleguide.
+   * @ignore
+   */
+  isShowingPlaceholder: PropTypes.bool,
+  /** The text to display for the more info button. */
+  isUserOnMobile: PropTypes.bool.isRequired,
```
red-runner: Are these all right? I think some descriptions were moved 📍

huvber: agreed

PR #795, opened 2019-11-29
WEB-725 Fix height responsiveness

red-runner:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-725)

### What are the **various** thing does this PR do?
* Fixes the responsive issues happening in the `Gallery` component.
* Makes active thumnails for the `Gallery` more visible; also, including arrows and close button.
* Adds a dark background for all images in the `Gallery` in case an image is not loaded. 
* Makes 

### Any other notes
yoyo

### DIFF
--- a/src/styles/semantic/definitions/lodgify-ui-components/gallery.less
+++ b/src/styles/semantic/definitions/lodgify-ui-components/gallery.less
@@ -12,21 +12,52 @@
         Gallery
 --------------------------*/
 
+/* Thumbnails */
+
+.gallery-thumbnail {
+  &:hover {
+    box-shadow: @galleryPicturesHoverBoxShadow;
+    transition: @galleryPicturesTansition;
+    cursor: pointer;
+  }
+}
+
+/* Modal */
+
 .modals {
   .modal {
     &.is-lightbox {
+      display: flex !important;
+      justify-content: center;
+      align-items: center;
+      background-color: @galleryContentBackgroundColor;
+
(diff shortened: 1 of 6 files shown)


src/styles/semantic/definitions/lodgify-ui-components/gallery.less
```diff
@@ -12,22 +12,53 @@
         Gallery
 --------------------------*/
 
+/* Thumbnails */
+
+.gallery-thumbnail {
+  &:hover {
+    box-shadow: @galleryPicturesHoverBoxShadow;
+    transition: @galleryPicturesTansition;
+    cursor: pointer;
```
huvber: I don't think the transition should be in a variable. it won't change base on template or user customization. And using the same transition in different place doesn't give us so much value either. (it's like to add `cusor: @galleryPicturecursor` for me)

src/styles/semantic/definitions/lodgify-ui-components/gallery.less
```diff
@@ -12,22 +12,53 @@
...
         overflow: hidden;
 
+        .image-gallery,
+        .image-gallery-slide-wrapper,
+        .image-gallery-swipe,
+        .image-galley-slides,
+        .image-galley-slide,
+        .image-gallery-image {
+          height: inherit !important;
+        }
```
huvber: what was this for?

PR #803, opened 2019-12-11
WEB-488: Avatar component and update Host Profile

huvber:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-488)

create a Avatar component and use that in HostProfile

### Any other notes
image
image

### DIFF
--- a/src/components/elements/Avatar/component.js
+++ b/src/components/elements/Avatar/component.js
@@ -25,7 +25,10 @@ Component.defaultProps = {
   image: null,
 };
 Component.propTypes = {
+  /** The firstname used for the placeholder. */
   firstname: string,
+  /** The image url to show as Avatar. */
   image: string,
+  /** The lastname used for the placeholder. */
   lastname: string,
 };
--- a/src/components/elements/Avatar/component.js
+++ b/src/components/elements/Avatar/component.js
@@ -6,13 +6,13 @@ import { testidFactory } from 'utils/testid';
 
 const testid = testidFactory('avatar');
 
-export const Component = ({ firstname, lastname, image }) => (
+export const Component = ({ firstName, lastName, image }) => (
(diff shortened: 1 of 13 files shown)


src/components/elements/Avatar/component.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      const image = current.find(testid('image'));
+
+      expect(image.length).toBe(0);
+    });
+    it('should have a text element', () => {
+      const text = current.find(testid('text'));
+
+      expect(text.length).toBe(1);
+    });
+    it('the text element should contain the first letter of both firstname and lastname', () => {
```
Jonnygwi: this should be in a describe block.
```
describe('the text element', () => {
  it('should contain the first letter of both firstName and lastName', () => {
    ...
  });
});

File: src/components/elements/Avatar/component.js [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+  lastname: ' ',
+  image: null,
+};
+Component.propTypes = {
+  /** The firstname used for the placeholder. */
+  firstname: string,
+  /** The image url to show as Avatar. */
+  image: string,
+  /** The lastname used for the placeholder. */
+  lastname: string,
```
Jonnygwi: both last name and first name are two words so we should use camel casing in the prop types and fix the prop descriptions.

src/styles/semantic/themes/default/lodgify-ui-components/avatar.variables
```diff
@@ -0,0 +1,7 @@
+/*----------------------
+    Block Placeholder
+-----------------------*/
+
+@avatar-size: 75px;
+@avatar-background-color: #c5c5c5;
+@avatar-font-size: 18px;
```
Jonnygwi: We should use em for font-size

PR #804, opened 2019-12-11
WEB-739 Implement StickyMenu

red-runner:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-739)

* Implements `StickyMenu`.
* Implements `useScroll` custom hook.

### Istanbul
* A small part of the component test has been commented using `Istanbul ignore`. Behavioural test for this particular case will be added in the future using cypress. 

### Behaviour
scroll-change-tab
scroll-on-click
side-scroll

### DIFF
--- a/src/components/collections/StickyMenu/component.js
+++ b/src/components/collections/StickyMenu/component.js
@@ -1,0 +1,118 @@
+import React, { useState, useRef } from 'react';
+import { PropTypes } from 'prop-types';
+import { Menu } from 'semantic-ui-react';
+
+import { ShowOn } from 'layout/ShowOn';
+import { Link } from 'elements/Link';
+import { Icon } from 'elements/Icon';
+import { FlexContainer } from 'layout/FlexContainer';
+import { testidFactory } from 'utils/testid';
+import { buildKeyFromStrings } from 'utils/build-key-from-strings';
+
+import { useScroll } from '../../../hooks/useScroll';
+
+import { RIGHT_END, LEFT_END } from './constants';
+import { getActiveOnScroll } from './utils/getActiveOnScroll';
+
+const TEST_ID_PREFIX = 'stickyMenu';
+
+const testid = testidFactory(TEST_ID_PREFIX);
+
+export const Component = ({ stickyMenuItems }) => {
+  const [activeItem, setActiveItem] = useState(stickyMenuItems[0].text);
+
+  const menuComponent = useRef();
+
+  const activeItemOnScroll = useScroll(
+    /* istanbul ignore next */
+    () => {
+      setActiveItem('');
+      return getActiveOnScroll(stickyMenuItems);
+    },
+    0
+  );
+
+  const scrollToComponentOnMenuClick = (location, text) => {
+    document.querySelector(location).scrollIntoView({
+      behavior: 'smooth',
+    });
+
+    window.history.pushState('', '', location);
+
+    setActiveItem(text);
+  };
+
+  const scrollMenuRight = () => {
+    menuComponent.current.scrollLeft = RIGHT_END;
+  };
+
+  const scrollMenuLeft = () => {
+    menuComponent.current.scrollLeft = LEFT_END;
+  };
+
+  return (
+    <FlexContainer
+      alignItems="center"
+      className="sticky-menu"
+      data-testid={testid()}
+      justifyContent="center"
+    >
+      <ShowOn mobile tablet>
+        <div
+          className="arrow left"
+          data-testid={testid('arrow-left')}
+          onClick={() => scrollMenuLeft()}
+        >
+          <Icon name="chevron left" />
+        </div>
+      </ShowOn>
+      <div
+        className="menu-wrapper"
+        data-testid={testid('menu')}
+        ref={menuComponent}
+      >
+        <Menu>
+          {stickyMenuItems.map(({ text, location }, index) => (
+            <Menu.Item
+              active={activeItem === text || activeItemOnScroll === text}
+              key={buildKeyFromStrings(text, index)}
+              name={text}
+              onClick={(event, data) =>
+                scrollToComponentOnMenuClick(location, data.name)
+              }
+            >
+              <Link>{text}</Link>
+            </Menu.Item>
+          ))}
+        </Menu>
+      </div>
+      <ShowOn mobile tablet>
+        <div
+          className="arrow right"
+          data-testid={testid('arrow-right')}
+          onClick={() => scrollMenuRight()}
+        >
+          <Icon name="chevron right" />
+        </div>
+      </ShowOn>
+    </FlexContainer>
+  );
+};
+
+Component.displayName = 'StickyMenu';
+
+Component.defaultProps = {
+  stickyMenuItems: [],
+};
+
+Component.propTypes = {
+  /** The sticky menu items to display. */
+  stickyMenuItems: PropTypes.arrayOf(
+    PropTypes.shape({
+      /** The location reference to be set as href. */
+      location: PropTypes.string.isRequired,
+      /** The text displayed as a menu item. */
+      text: PropTypes.string.isRequired,
+    })
+  ),
+};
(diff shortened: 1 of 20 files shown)


src/components/collections/StickyMenu/component.js
```diff
@@ -0,0 +1,114 @@
...
+
+import { RIGHT_END, LEFT_END } from './constants';
+import { getActiveOnScroll } from './utils/getActiveOnScroll';
+
+const TEST_ID_PREFIX = 'stickyMenu';
+
+const testid = testidFactory(TEST_ID_PREFIX);
+
+export const Component = ({ stickyMenuItems }) => {
+  const [activeItem, setActiveItem] = useState(stickyMenuItems[0].text);
```
huvber: if `stickyMenuItems` is empty array this will broke

src/components/collections/StickyMenu/component.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const menuComponent = useRef(null);
+
```
huvber: it is not necessary to pass `null`

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
```diff
@@ -0,0 +1,127 @@
...
+.sticky-menu {
+  position: sticky;
+  top: 0;
+  box-shadow: @stickyMenuBoxShadowY;
+  background-color: @white;
+
+  .menu-wrapper {
+    height: @stickyMenuHeight;
+    overflow-y: auto;
+    scrollbar-width: none;
```
huvber: this is supported only by firefox
image

https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
```diff
@@ -0,0 +1,127 @@
...
+    margin: 0;
+    box-shadow: none;
+    border: none;
+
+    &::-webkit-scrollbar {
+      display: none;
+    }
+
```
huvber: In edge you will still see the scrollbar. There are better way to hide the scrollbar imho.

this is ONE example I found very fast. 
http://jsfiddle.net/5GCsJ/954/

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
```diff
@@ -0,0 +1,127 @@
...
+  .arrow-right {
+    width: @stickyMenuArrowWidth;
+    height: @stickyMenuHeight;
+    position: relative;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    cursor: pointer;
+
+    &:after {
```
huvber: I would create another class since most of this rules are shared by both arrows. For the gradient I will but in a variable just the first parameter and define only that as custom. There are a lot of repeat rules that can be put in a shared class

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
```diff
@@ -0,0 +1,127 @@
...
+      width: 100%;
+      left: 0;
+      bottom: 0;
+      transition: @stickyMenuActiveItemTransition;
+    }
+
+    &.active {
+
+      .ui.button {
+        color: @stickyMenuActiveColor !important;
```
huvber: Why important?

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
```diff
@@ -0,0 +1,127 @@
...
+
+    }
+
+    &:before {
+      display: none;
+    }
+
+    .ui.button {
+      color: @stickyMenuInactiveColor !important;
+      text-transform: capitalize;
```
huvber: also here. We should avoid use important as much as possible

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
red-runner: Inheritance is preventing me to override styles.

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
red-runner: Inheritance is preventing me to override styles.

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
red-runner: This has been simplified.

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
Jonnygwi: "Inheritance is preventing me to override styles." isn't a reason. A style can be inherited and overwritten with a better selector and without the use of !important. If however we have an !important on this property on the selector then I guess this is the only option.

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
Jonnygwi: Yeah. We need better support for this. Please find a alternative approach. What @huvber recommended looks quick and easy. Try that.

src/components/collections/StickyMenu/component.js
```diff
@@ -0,0 +1,114 @@
...
+  const activeItemOnScroll = useScroll(
+    /* istanbul ignore next */
```
Jonnygwi: Why are we ignoring test coverage?

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
```diff
@@ -0,0 +1,127 @@
...
+    justify-content: center;
+    align-items: center;
+    cursor: pointer;
+
+    &:after {
+      content: '';
+      position: absolute;
+      width: 100%;
+      height: 100%;
+      background-color: #eeeeee;
```
Jonnygwi: Can we use a colour we already have that's a variable?

src/components/collections/StickyMenu/component.js
red-runner: Because we are trying to test behaviour that happens in the actual DOM, which in this case, would be best to test it with cypress. With andrea, we thought it was best to ignore this for now and add a test in the future.

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
red-runner: The inheritance of another !important is the reason. I added a tech debt card to have a look at the importants we have spread in our code and try to fix them.

src/components/collections/StickyMenu/component.js
huvber: I can answer that. This behavior is very difficult to test with unit-test. Is it almost impossible I would say because it's needed to take in consideration a lot of things. Mainly unit tests for frontend are using JSDom to simulate the DOM but they don't simulate/emulate/mock the browsers behaviour (window and document). Mocking them is a lot of effort right now (we tried for at least one hour together) and we concluded that is not worthy in this moment and for the kind of feature. This can be easily test and covered by integration test (aka Cypress) that we are not still using but we should as soon as we reach a stable status of our project.

src/styles/semantic/definitions/lodgify-ui-components/sticky-menu.less
red-runner: It is a reason, we are inheriting !important from other components. I created a tech debt card to fix this things in the future

src/components/collections/StickyMenu/component.js
huvber:
```
const [activeItem, setActiveItem] = useState('');

useEffect(() => {
     if (stickyMenuItems.length > 0) {
         setActiveItem(stickyMenuItmes[0].text);
     }
},[stickyMenuItems.length]);
```
```

PR #815, opened 2020-01-08
WEB-734 Contact empties all inputs after submission

red-runner:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-734)

Makes all the `Contact` inputs empty after submission.

### Any other notes

* Discovered an issue where, after submission, recaptcha is not refreshed causing the form to error on a second submission; even if all fields are correct.
[Related Jira Issue](https://lodgify.atlassian.net/browse/WEB-856?atlOrigin=eyJpIjoiMTQ3NmQ3ZjhiZmNkNGI1OWJjZDQ5ODljNWQ3NWIxODMiLCJwIjoiaiJ9)

asdasd

### DIFF
--- a/src/components/collections/Form/utils/getEmptyState.js
+++ b/src/components/collections/Form/utils/getEmptyState.js
@@ -6,9 +6,7 @@ export const getEmptyState = state =>
   Object.keys(state).reduce(
     (accumulator, key) => ({
       ...accumulator,
-      [key]: {
-        value: null,
-      },
+      [key]: { value: key === 'dates' ? null : '' },
     }),
     {}
   );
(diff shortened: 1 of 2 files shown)


src/components/collections/Form/utils/getEmptyState.js
```diff
@@ -6,9 +6,12 @@ export const getEmptyState = state =>
...
-      [key]: {
-        value: null,
-      },
+      [key]:
+        key === 'dates'
+          ? {
+              value: null,
+            }
+          : { value: '' },
     }),
```
huvber: I think it will be more clean to write:
```
[key]: {
   value: key === 'dates' ? null : ''
}
```

PR #823, opened 2020-01-14
WEB-787: refactor responsiveImage and align other components.

huvber:
[Related Jira issue](https://lodgify.atlassian.net/browse/WEB-787)

refactor Image to align all component with the same behavior.

### Any other notes
This PR will simplify the ResponsiveImage Component and align all the other images inside the library to have the same behavior and be consistent.
image


src/styles/semantic/themes/livingstone/elements/image.overrides
```diff
@@ -18,83 +18,50 @@
...
+
+  .ui.image {
+    visibility: hidden;
+    filter: blur(20px);
+    scale: scale(1.15);
+    transition: 0.25s ease-out;
   }
 
-  &.has-blurred-children {
+  &.isLoaded {
```
Jonnygwi: I don't think we use camel case for css classes. Let's stick to kebab for consistency

src/styles/semantic/themes/livingstone/elements/image.overrides
```diff
@@ -18,83 +18,50 @@
...
-      transform: @imageWithPlaceholderTransformScale;
+    .ui.image {
+      visibility: visible;
+      filter: none;
+      scale: none;
     }
   }
 
-  &.has-placeholder {
+  &.isCircular {
```
Jonnygwi: Same for this one https://github.com/lodgify/lodgify-ui/pull/823/files?file-filters%5B%5D=.config&file-filters%5B%5D=.js&file-filters%5B%5D=.less&file-filters%5B%5D=.md&file-filters%5B%5D=.overrides&file-filters%5B%5D=.variables#r366268699

src/styles/semantic/themes/livingstone/elements/image.overrides
```diff
@@ -18,83 +18,50 @@
...
+      position: absolute;
       top: 0;
+      bottom: 0;
       width: 100%;
+      height: 100%;
     }
   }
 
-  &.is-fluid {
+  &.willFill {
```
Jonnygwi: https://github.com/lodgify/lodgify-ui/pull/823/files?file-filters%5B%5D=.config&file-filters%5B%5D=.js&file-filters%5B%5D=.less&file-filters%5B%5D=.md&file-filters%5B%5D=.overrides&file-filters%5B%5D=.variables#r366268699

src/components/media/Slideshow/component.js
```diff
@@ -137,4 +138,6 @@ Component.propTypes = {
   isStretched: PropTypes.bool,
   /** The index used to define the image shown when the slideshow is opened. */
   startIndex: PropTypes.number,
+  /** Whether to render the image that fill the container. */
```
Jonnygwi:
```suggestion
  /** The rendered image will fill the container */
```

src/styles/semantic/themes/livingstone/elements/image.overrides
```diff
@@ -18,83 +18,50 @@
...
     height: 100%;
-    width: 100%;
-
-    /* Object fit cover Fallback */
```
Jonnygwi: Why have you removed the fallback when you're still using object-fit?

src/styles/semantic/themes/livingstone/elements/image.overrides
huvber: I used camel case because this simplify a lot the code and make it easier to find out which javascript flag will control specific css properties. It's also consistent with the style of classes used in the other projects. I'm asking you to take this reasons in consideration and then is up to you and @red-runner to decided to stick to the old style without this advantages or starting that. I don't have real problem to convert that to kebab case but I don't find any good advantage for that.

src/styles/semantic/themes/livingstone/elements/image.overrides
huvber: I used another flag because I'm afraid this will create collision with styles used in semantic-ui or in other classes.

src/styles/semantic/themes/livingstone/elements/image.overrides
huvber: I didn't use object-fit and actually that wasn't a fallback since was applied also when object-fit is compatible

Jonnygwi: Well. The styles get overridden then object-fit is compatible but okay.

Jonnygwi: https://github.com/lodgify/lodgify-ui/pull/823/files/0d73e05af8bb35163234f2d20e2df72c755e0df0#diff-0b263866575efd64b3c073a45fc6d837R50

src/styles/semantic/themes/livingstone/elements/image.overrides
Jonnygwi: Okay well I'm only in the consistency boat so if you want to do this then open a tech debt ticket to convert all the others at a later date. Also make sure @red-runner is aware of this coding style change.

src/styles/semantic/themes/livingstone/elements/image.overrides
Jonnygwi: I think we should test that to make sure it doesn't conflict rather than adding `willFill` as a prop. Having two props `willFill` and `isFluid` for one component is going to confuse anyone trying to use it. 🤯

src/styles/semantic/themes/livingstone/elements/image.overrides
huvber: I forgot I added then later sorry. If you want I will try to restore the object-fit fallback even if only ie10 is not compatible with that and we use much more destructive features not compatible with ie10 without put the any fallback.

